PR #1855, opened 2026-04-18
fix(orgchart): guard same-parent drops + confirm reset

HankHuang0516:
## Summary
Fixes two P2 org-chart web bugs identified during the Android embed E2E drill (kanban cards for drag-dangling-child and reset-no-confirm).

- **Drag dangling child**: dropping a node onto its current direct parent is now a no-op. Previously `moveInHierarchy` would remove-then-re-add under the same parent, which opens a transient window where the entity is briefly absent from the hierarchy and `drawConnectors` / `nodeEls` can race, occasionally rendering a child DOM node with no incoming SVG connector. Guarded in both the mouse drop handler and the touch drop handler.
- **Reset to Default**: now shows a destructive `showConfirm` dialog before flattening the hierarchy. Matches the pattern already used for cross-device reset (`dash_xd_reset_confirm`) and for per-entity clear (`dash_id_clear_confirm`).
- New i18n key `org_reset_confirm`: en / zh-TW / zh-CN translated; other locales inherit English via `i18n.t` fallback. Separate i18n card will route the remaining 12+ locales via speakTo.

## Test plan
- [ ] Android WebView (com.hank.clawlive ≥ 1.0.75): open Org Chart sheet → touch-drag a sibling onto its current parent → expect no visual change, no save.
- [ ] Android WebView: touch-drag a sibling onto a different sibling → expect reparent + connector drawn correctly (no dangling child).
- [ ] Desktop: mouse-drag a child onto its current parent → no-op.
- [ ] Reset to Default → expect confirm dialog → Cancel keeps hierarchy intact → OK flattens.
- [ ] Verify zh-TW / zh-CN locale shows translated confirm text.

@claude review


HankHuang0516:
@claude please review this PR for correctness + coverage (web orgchart drag dangling-child guard + reset confirm dialog)

PR #1868, opened 2026-04-19
fix(allocator): tombstone freed publicCodes — prevent stale-QR re-routing

HankHuang0516:
## Motivation

Fixes P2 from the allocator audit (`card_0256a6c043d20be35b8388d1`).

`generatePublicCode()` only checked the live `publicCodeIndex`, so a 6-char code freed by an unbind / permanent-delete could be re-handed-out to a different entity on the next call. Anyone holding a stale reference (saved agent card, QR code on a fridge, channel-account record, mention token, /p/<code> bookmark) would then be silently re-routed to the wrong entity.

With ~36^6 = 2.1B keyspace and the platform's growing churn (each unbind frees a code), the collision probability is non-negligible over time — and unlike a normal collision, the failure mode is silent (no error, just wrong destination).

## Approach: tombstone Set (option 1 of the audit's 3 suggestions)

The audit suggested option 3 (`pc_<n-hex>` prefix scheme), but that's blocked by the 6-char-format assumption hard-coded in 4+ places: the validator regex `^[a-z0-9]{6}$` (index.js:10156), the mention parser regex `<@([a-z0-9]{6})>` (`mention-parser.js`), the client-side `mention-render.js`, and equivalent regexes on Android/iOS. The PR constraint was index.js + a single jest test, so the format change is out of scope.

Option 1 (tombstone Set) is contained, deterministic, and piggybacks on the existing `entity_trash` table for cross-restart persistence (no schema change).

## Implementation

- Wrap `publicCodeIndex` in a Proxy whose `deleteProperty` trap auto-records every released key into `deletedPublicCodes` (a Set). This catches every existing free site (unbind, permanent-delete, custom-set rename, `rental.js` helper deletes, `channel-api.js` cleanup) without threading a helper through every module.
- `generatePublicCode()` rejects any candidate already in `publicCodeIndex` OR in `deletedPublicCodes`. Retry budget bumped 10 → 20 to compensate for the slightly higher collision pressure.
- New `loadTombstonesFromTrash()` seeds the set on startup from `entity_trash` (existing table, 7-day retention) so freed codes survive restart for the soft-delete window. Skips any code currently held by a live entity (live wins).
- **Ordering** in `initPersistence()`: `buildPublicCodeIndex` → `loadTombstonesFromTrash` → `backfillPublicCodes`. This ordering matters — without it, a backfilled code could accidentally equal a still-trashed code (caught in coverage review).
- Trash-restore site (`/api/device/entity-trash/:trashId/restore`) and custom-set rename site (`/api/device/restore-entity-public-code`) explicitly clear the tombstone for the code they're re-installing.
- New module exports `_generatePublicCode` / `_publicCodeIndex` / `_deletedPublicCodes` for the test (matches the existing `_sanitizePublicHtml` pattern).

## Constraint compliance

- ✅ `backend/index.js` only (+ a new jest test). No schema change. No migration script for existing codes. No format change. Old 6-char codes continue to work everywhere.

## Test plan

- [x] `backend/tests/jest/public-code-allocator.test.js` — 7 cases incl. force-collide (mock `crypto.randomBytes` to feed 19 already-tombstoned candidates and assert generator skips them) and a 1k generate→delete→generate cycle that asserts no overlap with the tombstone Set.
- [x] Full Jest suite (92 suites, 1602 tests) passes locally.
- [x] `npx eslint backend/index.js` clean.
- [ ] Verify on production after merge: `/api/health` deploy canary; spot-check that an unbind followed by a fresh bind doesn't recycle the same code (manual smoke).

## Notes for reviewers

- The Proxy is only on `publicCodeIndex` — the underlying `_publicCodeStore` is purely an implementation detail. Reads, writes, `Object.keys()`, and `in` checks all behave identically to the prior plain-object form (verified — no caller iterates the index).
- Tombstone set is in-memory but seeded from `entity_trash` on every startup. After a code's trash row expires (7-day retention) and the server restarts, that code becomes re-issuable again. This is an acceptable trade-off for not adding a permanent tombstones table.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -872,10 +872,12 @@ async function initPersistence() {
         }
     }
 
-    // Build public code index and backfill any missing codes
+    // Build public code index, then seed tombstones from trash BEFORE backfill
+    // so a freshly generated backfill code can never accidentally collide with
+    // a still-referenced (but trashed) public code.
     buildPublicCodeIndex();
-    await backfillPublicCodes();
     await loadTombstonesFromTrash();
+    await backfillPublicCodes();
 
     // Load DB-approved skill contributions (supplements git-tracked skill-templates.json)
     if (usePostgreSQL) await loadApprovedContributions();
--- a/backend/index.js
+++ b/backend/index.js
@@ -654,7 +654,24 @@ const BOT2BOT_REGEN_INTERVAL_MS = 60 * 1000; // regenerate 1 token every 60 seco
 const recentBroadcasts = {}; // deviceId -> [{fromEntityId, text, timestamp}] for dedup
 
 // Cross-device messaging
-const publicCodeIndex = {}; // code -> { deviceId, entityId }
+//
+// publicCodeIndex maps a 6-char public code → { deviceId, entityId }.
+// We wrap the underlying object in a Proxy whose `deleteProperty` trap
+// records every released code into `deletedPublicCodes` (a tombstone Set).
+// generatePublicCode() then refuses to re-issue any tombstoned code, which
+// prevents stale QR codes / saved cards from being silently routed to a
+// different entity that happened to grab the same 6 chars later (FK risk
+// flagged in the allocator audit, card_0256a6c043d20be35b8388d1).
+const _publicCodeStore = {};
+const deletedPublicCodes = new Set();
+const publicCodeIndex = new Proxy(_publicCodeStore, {
+    deleteProperty(target, prop) {
+        if (typeof prop === 'string' && prop in target) {
+            deletedPublicCodes.add(prop);
+        }
+        return delete target[prop];
+    },
+});
 const crossSpeakCounter = {};
 const crossDeviceOwnerRateLimit = {}; // owner-defined per-sender rate limit timestamps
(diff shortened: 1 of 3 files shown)


HankHuang0516:
Follow-up commit cc735940 (`harden(1868)`) addresses coverage-review nit #1: all four `rental.js` sites that delete publicCodes (`removeRentalEntity`, `reconcileRentalEntities` phase 1 + phase 2, `debug/cleanup-rental-ghosts`) now call `saveToEntityTrash` BEFORE the `delete publicCodeIndex[code]`, matching the index.js pattern at L5655/5773/6113. Wired `saveToEntityTrash` into both `reconcileRentalEntities` helpers (~L1841) and `setInterviewDeps` (~L13156). Restart-survival guarantee now holds on rental paths.

Nits #2 / #3 intentionally skipped: both target helpers (`loadTombstonesFromTrash`, `generatePublicCode`) live inside the monolithic `index.js` and are not module-exported — unit-testing them would require a refactor beyond this hardening follow-up.

All 6 rental/entity-trash jest files still pass (93 tests, 0 failures). CI green before merge.

PR #1879, opened 2026-04-19
feat(wallet): verify Google Play purchases via androidpublisher v3

HankHuang0516:
## Summary

Wires `/api/wallet/topup/verify-google` to Google Play's androidpublisher v3 REST API before crediting the wallet. Replaces the `TODO: validate purchaseToken` trust-token path at `backend/wallet.js:794` with a fully verified pipeline.

- `verifyGooglePurchase(packageName, productId, purchaseToken)` — service-account JWT → OAuth token → GET purchase-token resource. Typed errors: `google_auth_failed`, `google_token_invalid`, `google_token_canceled`, `google_token_pending`.
- `acknowledgeGooglePurchase()` — fire-and-forget `:acknowledge` call after the ledger credit succeeds. Ack failures log a warn audit but never revoke credited ecoins (a retry cron can reconcile; within Google's 3-day ack window).
- Route uses Google's **stable `orderId`** as `external_txn_id`, not the raw purchaseToken — `UNIQUE(channel, external_txn_id)` still dedupes across the full lifecycle.
- Singleton GSA env parser + 55-minute access-token cache (conservative under Google's 60-min TTL) with single-flight `inflight` tracking to avoid thundering-herd re-signs on first use.

## Security posture — FAIL-CLOSED

Missing / unparseable `GOOGLE_PLAY_SERVICE_ACCOUNT` now rejects requests with `500 google_auth_failed` instead of the legacy trust-token fallback. Rationale: this endpoint credits real money; an env misconfiguration should not open the door to forged purchaseTokens.

If Hank wants the fallback behavior for a production rollback, set `GOOGLE_PLAY_ALLOW_UNVERIFIED=1` on Railway — the flip is explicit, audited (warn-level), and temporary. Unset it to re-enable verification.

Other guardrails:
- **No GSA leakage:** private key / full JSON never logged. Audit records only include `client_email` / `project_id`.
- **Constant-time checks:** existing `safeEqual` preserved for device auth; no new cmp paths added.
- **Input masking:** upstream auth failure reasons (e.g. `invalid_grant`) are kept in server audit logs only, never surfaced to the client response body.

## Changes

- `backend/wallet.js`
  - New constants `GOOGLE_PACKAGE_NAME`, `GOOGLE_TOKEN_CACHE_MS`, OAuth / androidpublisher URL literals.
  - New helpers (above the Express factory): `_getServiceAccount()` singleton, `_signServiceAccountJwt()`, `_getGoogleAccessToken()` with single-flight cache, `verifyGooglePurchase()`, `acknowledgeGooglePurchase()`, `_resetGoogleCache()` (test-only).
  - Route updated to call `verifyGooglePurchase` → map typed errors to `402 purchase_not_verified` (client-facing) or `500 google_auth_failed` (server-side). Fire-and-forget ack post-credit.
  - Uses `Google orderId` for `external_txn_id`; `externalRaw` now stamps `{ productId, source, orderId, verified }` for forensic.
  - Factory + module-level exports: `verifyGooglePurchase`, `acknowledgeGooglePurchase`, `GOOGLE_PACKAGE_NAME`, `_resetGoogleCache`.

- `backend/tests/jest/wallet-topup-google.test.js` (new)
  - Generates a throwaway RSA keypair in-memory so the JWT signer runs for real (never a checked-in secret).
  - Mocks `global.fetch` with URL-dispatched responders (token / products GET / ack) and asserts call counts.
  - 15 test cases covering all 8 spec scenarios plus input validation:
    1. Happy path → `{valid:true, orderId}` and credit keyed by orderId
    2. `purchaseState=1` canceled → 402
    3. `purchaseState=2` pending → 402
    4. Token HTTP 401 → 500 (detail masked)
    5. Unset GSA → 500 fail-closed
    5b. `GOOGLE_PLAY_ALLOW_UNVERIFIED=1` → audited rollback path
    6. Token cache single-flight (1 token POST for 2 verifies)
    7. Dedup via orderId (second call `deduped:true`, balance credited once)
    8. Ack rejects → endpoint still 200, ledger still credited

- `backend/tests/jest/wallet.test.js`
  - Existing `POST /topup/verify-google credits wallet end-to-end` test now uses the `GOOGLE_PLAY_ALLOW_UNVERIFIED=1` rollback so it doesn't need Google creds. Full Google-API path coverage lives in the new file.

## Tradeoffs

- **Fail-closed over fallback-to-trust:** spec allowed either; I picked fail-closed because the endpoint credits money and env misconfigurations should never be silently turned into forged-token acceptance. Escape hatch: `GOOGLE_PLAY_ALLOW_UNVERIFIED=1`.
- **No new deps:** uses `jsonwebtoken` (already in `backend/package.json`) and native `fetch` (Node 18+).
- **Ack is best-effort:** never blocks the response and never throws. If Google's ack endpoint flaps, users keep their ecoins — a retry cron is the correct compensation path (ack is required within 3 days; we have a large window).
- **Token cache race:** single-flight via `_accessTokenCache.inflight` — concurrent callers on cold-start await the same promise instead of re-signing.

## Test plan

- [x] `npx jest tests/jest/wallet-topup-google.test.js` → 15/15 pass
- [x] `npx jest tests/jest/wallet.test.js` → 55/55 pass (existing case updated to exercise rollback)
- [x] Full `npx jest` backend suite → 1673/1673 pass across 96 suites
- [ ] Live smoke post-merge: fake token → 402; real TWD 30 tier token → credits 3000 e幣 exactly once (dedup check)
- [ ] Verify Railway env `GOOGLE_PLAY_SERVICE_ACCOUNT` is set + parseable before the first real user attempt (fail-closed will reject otherwise)

### DIFF
--- a/backend/tests/jest/wallet-topup-google.test.js
+++ b/backend/tests/jest/wallet-topup-google.test.js
@@ -1,0 +1,515 @@
+/**
+ * Google Play androidpublisher v3 verification tests.
+ *
+ * Covers the 8 cases from the spec:
+ *   1. Happy path: purchaseState=0 → verifyGooglePurchase returns
+ *      { valid:true, orderId }; endpoint credits the ledger using
+ *      Google's orderId (NOT the raw purchaseToken) as external_txn_id.
+ *   2. purchaseState=1 (canceled) → 402 purchase_not_verified.
+ *   3. purchaseState=2 (pending)  → 402 purchase_not_verified.
+ *   4. Token exchange HTTP 401   → 500 google_auth_failed (detail masked).
+ *   5. GOOGLE_PLAY_SERVICE_ACCOUNT unset → FAIL-CLOSED 500 google_auth_failed.
+ *      (We chose fail-closed over fallback-to-trust; see PR body.)
+ *   6. Access-token cache: two successive verifications → only ONE POST
+ *      to oauth2.googleapis.com/token.
+ *   7. Dedup: same purchaseToken credited twice → second call is deduped
+ *      (same orderId in response; wallet balance unchanged).
+ *   8. Acknowledge is fire-and-forget: mock ack to reject → endpoint
+ *      still returns 200 and ledger is still credited.
+ *
+ * Strategy: reuse the wallet.test.js pg simulator (same jest.mock shape),
+ * mock `global.fetch` per test, and drive the route via supertest.
+ */
+
+// ── In-memory pg simulator (mirrors wallet.test.js) ─────────────────────
+jest.mock('pg', () => {
+    const state = {
+        wallets: new Map(),
+        ledger: [],
+        nextLedgerId: 1,
+        topupOrders: [],
+        nextOrderId: 1,
+        userAccounts: new Map(),
+    };
+    globalThis.__walletFakeState = state;
+
+    function ensureWallet(userId) {
+        if (!state.wallets.has(userId)) {
+            state.wallets.set(userId, {
+                balance_mli: 0n,
+                held_mli: 0n,
+                lifetime_earned_mli: 0n,
+                lifetime_spent_mli: 0n,
+            });
+        }
+    }
+
+    function runQuery(sql, params = []) {
+        const norm = sql.replace(/\s+/g, ' ').trim();
+        if (/^(BEGIN|COMMIT|ROLLBACK)$/i.test(norm)) return { rows: [], rowCount: 0 };
+        if (/^CREATE (TABLE|INDEX|UNIQUE INDEX)/i.test(norm)) return { rows: [], rowCount: 0 };
+        if (/^INSERT INTO wallets \(user_id\) VALUES/i.test(norm)) {
+            ensureWallet(params[0]);
+            return { rows: [], rowCount: 1 };
+        }
+        if (/^SELECT .* FROM wallet_ledger WHERE idempotency_key = \$1$/i.test(norm)) {
+            const match = state.ledger.find(r => r.idempotency_key === params[0]);
+            return { rows: match ? [match] : [], rowCount: match ? 1 : 0 };
+        }
+        if (/^SELECT balance_mli, held_mli FROM wallets WHERE user_id = \$1 FOR UPDATE$/i.test(norm)) {
+            const w = state.wallets.get(params[0]);
+            if (!w) return { rows: [], rowCount: 0 };
+            return {
+                rows: [{ balance_mli: w.balance_mli.toString(), held_mli: w.held_mli.toString() }],
+                rowCount: 1,
+            };
+        }
+        if (/^UPDATE wallets SET balance_mli = \$1, held_mli = \$2/i.test(norm)) {
+            const [bal, held, earnedInc, spentInc, userId] = params;
+            const w = state.wallets.get(userId);
+            if (!w) return { rows: [], rowCount: 0 };
+            w.balance_mli = BigInt(bal);
+            w.held_mli = BigInt(held);
+            w.lifetime_earned_mli += BigInt(earnedInc);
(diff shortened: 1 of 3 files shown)


HankHuang0516:
## Coverage Review (adversarial)

Reviewed the diff in-process (Agent-tool `general-purpose` spawn not available from this harness; performed the same checklist inline). Verdict: **no P0/P1 issues — safe to merge.** One **P2 follow-up** noted at the bottom.

### 1. Can an attacker forge a valid purchaseToken?
**No.** The token is round-tripped through Google:
- Our service-account JWT is signed with `sa.private_key` (only Railway has it) and exchanged at `oauth2.googleapis.com/token` over TLS.
- The returned access token authorizes a GET against `androidpublisher.googleapis.com/v3/applications/com.eclawbot.app/purchases/products/{productId}/tokens/{purchaseToken}`.
- Unknown tokens → HTTP 400/404/410 → `google_token_invalid` → 402 `purchase_not_verified`.
- We use Google's stable `orderId` (not the raw token) as `external_txn_id`, so even two users submitting the same authentic token only credit once via `UNIQUE(channel, external_txn_id)`.

### 2. Token-cache race on cold start
**Safe.** Single-flight via `_accessTokenCache.inflight`:
- First caller sets `inflight` to the promise.
- Concurrent callers at `wallet.js:837-839` return the same promise.
- `finally` clears `inflight` after settle; post-settle callers see populated `token/expiresAt` and hit the fast path.
- Failures don't poison the cache (token/expiresAt only written on success) — correct behavior.

### 3. Ack-failure compensation
**Correct, with P2 note.** The ack is fire-and-forget:
- `Promise.resolve().then(() => acknowledgeGooglePurchase(...))` runs after the HTTP response is sent.
- Failures `audit('warn', ...)` but never revoke credited ecoins — matches the spec's "ack failure is recoverable via retry cron; never leave a refunded-but-credited gap."
- **P2 gap:** no retry cron is implemented in this PR. If Google's ack endpoint flaps and the 3-day window lapses, Google auto-refunds the user but our ledger keeps the credit → free top-up (revenue leak, not a security vuln). Recommend follow-up card: periodic sweep over `topup_orders WHERE channel='google_play' AND paid_at < now-12h AND externalRaw->>'verified'='true'` → call `acknowledgeGooglePurchase` for any that Google still reports as `acknowledgementState=0`.

### 4. GSA secret leaking in logs
**Clean.** Audited every log site:
- `_getServiceAccount` / `_signServiceAccountJwt` / `_getGoogleAccessToken`: zero log statements.
- Route `audit('error', ..., reason=${vErr.reason})` only logs sanitized reason codes (`gsa_missing_or_invalid`, `token_exchange_http_401`, `api_http_403`) — never `private_key`, `private_key_id`, or raw JSON.
- `console.error('[Wallet] topup verify-google error:', err)` catch-all: errors we throw carry only `.reason` (strings) and `.purchaseState` (ints). No GSA leakage.
- `externalRaw` stored in DB: `{ productId, source, orderId, verified }` — no secrets.

### 5. Timing attacks
No new comparison paths. Device-auth still uses existing `safeEqual`. JWT validation is Google's responsibility (not ours), and OAuth token matching is not done on our side.

### 6. Minor observations (non-blocking)
- `GOOGLE_TOKEN_CACHE_MS=55min` is a hard-coded conservative of Google's 3600s tokens. If Google ever shortens the TTL, we could use a stale token for up to 55 min before seeing a 401. Low risk in practice; worst case → one user sees `google_auth_failed` until cache expires. Acceptable.
- `audit('warn' ...)` on unverified fallback (GOOGLE_PLAY_ALLOW_UNVERIFIED=1) fires on **every** request, not once — this is intentional so Hank sees it in production log streams.

### Verdict
**PASS — proceeding to merge.** Follow-up card for the ack-retry cron will be filed separately (P2, revenue-safety not security).

PR #1919, opened 2026-04-20
fix(chat): strip ↩ quote prefix to prevent nested reply quotes

HankHuang0516:
## Summary
- Fix reply snippet/preview truncating at the wrong boundary when the source message was itself a reply (prefixed with `↩ <sender>: "<excerpt>"\n\n<body>`).
- Introduces `stripReplyQuotePrefix()` helper; applied in the chat-reply-btn `data-text` generation (chat.html:3920) and in `setReplyContext()` as defense-in-depth.

## Why
Before: replying to an already-replied message produced nested `↩ 你: "↩ 你: \"…\""` because the 120-char snippet consumed the old quote header instead of the actual body.
After: the preview and the outgoing quote excerpt always show the real message body.

## Test plan
- [ ] In `/portal/chat.html`, send message A. Reply to A with B. Reply to B with C.
- [ ] Verify B's reply button preview shows A's body (not the "↩" wrapper).
- [ ] Verify C arrives as `↩ 你: "B's actual body"\n\nC-text`, not nested `↩ 你: "↩ 你: …"`.
- [ ] Non-reply messages still show their natural 120-char snippet.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,6 +30,8 @@ EClaw/
 │   ├── device-telemetry.js   # AI debug buffer per device
 │   ├── device-feedback.js    # Feedback/bug report system
 │   ├── chat-integrity.js     # Chat message integrity validation
+│   ├── chat-embedding.js     # pgvector schema + cosine search + embedMessageAsync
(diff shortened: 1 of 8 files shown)


HankHuang0516:
@claude please review coverage

PR #1924, opened 2026-04-21
feat(kanban): sort dropdown on portal board

HankHuang0516:
## Summary
- Adds a 4-option sort select to the kanban portal toolbar: **Default Order / Newest First / Oldest First / High Priority First**.
- Sort runs per-column **after** the existing All/My-Cards filter, so filtering still works.
- New module-level `currentSort` state + `sortCards()` / `applySorting()` helpers; `renderBoard()` now calls `applySorting()` on each column's filtered cards.

## Test plan
- [ ] Open `/portal/kanban.html`, confirm dropdown renders next to All/My-Cards chips.
- [ ] Flip each of the 4 options, verify cards in each column visibly reorder.
- [ ] Combine with `My Cards` filter — sort must still apply.
- [ ] Check browser console: no errors, no `NaN` sort warnings.
- [ ] Mobile list view still renders (not yet sort-aware — flagged for follow-up).

## Known gaps (follow-up card)
- Sort option labels are hardcoded EN. Needs `kb_sort_*` i18n keys across all 5 locales in `backend/public/shared/i18n.js`.
- `renderMobileList()` not yet plugged into `applySorting()`.

### DIFF
--- a/backend/public/portal/kanban.html
+++ b/backend/public/portal/kanban.html
@@ -938,7 +938,7 @@
                 case 'priority_asc':
                     return sorted.sort((a, b) => {
                         const pOrder = { 'P0': 0, 'P1': 1, 'P2': 2, 'P3': 3 };
-                        return (pOrder[a.priority] || 999) - (pOrder[b.priority] || 999);
+                        return (pOrder[a.priority] ?? 999) - (pOrder[b.priority] ?? 999);
                     });
                 default:
                     return cards;
--- a/backend/public/portal/kanban.html
+++ b/backend/public/portal/kanban.html
@@ -23,6 +23,7 @@
         .kb-chip { padding:4px 12px; border-radius:var(--radius-pill); font-size:12px; font-weight:500; cursor:pointer; border:1px solid var(--card-border); background:none; color:var(--text-secondary); transition:all 0.2s; }
         .kb-chip.active { background:var(--primary); color:#fff; border-color:var(--primary); }
         .kb-chip:hover:not(.active) { border-color:var(--text-muted); color:var(--text); }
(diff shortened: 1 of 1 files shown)


HankHuang0516:
@claude please review coverage for this PR (kanban sort dropdown). Focus areas:
(1) applySorting() correctness — createdAt fallback to 0, priority order P0→P3 with 999 fallback, stable default order
(2) sortCards() + currentSort state interaction with existing filterCards() — does filter+sort compose correctly?
(3) renderBoard() integration — per-column sort inside STATUSES.forEach still matches existing count/render semantics
(4) mobile list view (renderMobileList) — is sort silently broken there? flagged in PR body as follow-up
(5) any XSS / DOM injection risk from new sort select

HankHuang0516:
## Self-review (as Claude, #2)

**Coverage**: HTML (1 CSS rule + 5-line <select>), state (1 var), renderBoard integration (1 line split), sortCards/applySorting (~25 lines). E2E harness already validated all 4 sort modes yesterday before commit.

**Correctness**:
- Line 941 `pOrder[a.priority] ?? 999` uses nullish coalescing — correctly avoids the P0 falsy-trap (0 || 999 === 999 would have sent P0 last).
- `applySorting` takes a fresh copy via `[...cards]` so the original `filtered` array isn't mutated across columns.
- `sortCards` → `renderBoard(true)` matches existing `filterCards` flow.
- Default case returns untouched cards → safe fallback.

**Compose with filter**: filter runs first (line 815), sort runs inside per-status forEach (line 816) — counts and render both operate on the sorted array, so UI stays consistent.

**XSS**: sort `<select>` has hardcoded options with no user-provided values; `onchange` binds to literal `sortCards(this.value)` where value is one of 4 whitelisted strings that feed into a switch — nothing reaches innerHTML.

**Not covered (intentional, out of scope)**:
- mobile list view (`renderMobileList`) — flagged in PR body as follow-up
- i18n of sort option labels — moved to 5 per-locale cards (Mac_E) + 1 wiring card (#2) per 1-locale-per-PR granularity rule

LGTM. Merging.

PR #1918, opened 2026-04-20
feat(chat): semantic vector search via pgvector + OpenAI embeddings (P1)

HankHuang0516:
## Summary

Adds end-to-end infrastructure for searching chat history by meaning via pgvector. Ships in shadow mode: if pgvector is unavailable or no embedding API key is configured, `/api/chat/search` transparently falls back to ILIKE keyword search — zero crash surface added to the chat write path.

### What's new
- **`chat-embedding.js`** — `initSchema` (CREATE EXTENSION vector → ALTER TABLE ADD COLUMN embedding vector(1536) → HNSW index); `embedMessageAsync` fire-and-forget; `searchBySemantic` (cosine `<=>`) + `searchByKeyword` fallback.
- **`embedding-client.js`** — OpenAI text-embedding-3-small (1536) default, Voyage-3 optional via `CHAT_EMBEDDING_PROVIDER=voyage`. BYO key via device-vars (`OPENAI_API_KEY` / `VOYAGE_API_KEY`); server env is fallback.
- **`saveChatMessage` hook** — fires `embedMessageAsync` after every successful INSERT; no blocking on response.
- **`POST /api/chat/search`** — dual-mode endpoint. Returns `mode: "semantic"` / `"keyword"` / `"semantic_empty_keyword_fallback"` so the caller can detect which path ran. Auth via `deviceSecret` or `botSecret`. `entityId` optional (bot-auth tokens are always scoped to their own entity regardless).
- **`scripts/backfill-embeddings.js`** — idempotent batch backfill. `DATABASE_URL=... OPENAI_API_KEY=... node backfill-embeddings.js --batch=50 --max=2000 --device=X --dry-run`.
- **`.env.example`** — documents new env vars.
- **Jest `chat-search.test.js`** — validation, auth, soft-fail, helper modules; 13 new tests covering the disabled-pipeline path (CI never hits OpenAI).
- **CLAUDE.md** — route table + module list.

## Rollout phases
1. **Schema migration** (this PR) — soft-fail if pgvector absent on the managed PG
2. **Shadow embedding** (this PR, fire-and-forget on write)
3. **`/api/chat/search` endpoint** (this PR)
4. Frontend integration — separate card, out of scope

## Test plan
- [x] Jest 1864/1864 local (includes 13 new chat-search cases)
- [x] ESLint clean on all new files
- [ ] Prod smoke: POST /api/chat/search with a real device — confirm endpoint responds 200 with `mode` field
[...]

PR #1926, opened 2026-04-21
feat(chat): /api/chat/message/:id/related — nearest-neighbor by anchor message

HankHuang0516:
## Summary
Completes PR #1918 follow-up "quote_ref 矩陣關聯" (card `card_f98be9deb1df16d23ed9e8ee`). Given a chat message id, returns the N messages whose embedding is closest to the anchor's embedding, excluding self. Falls back to ILIKE on the anchor's text when pgvector is off or the anchor has no embedding — same shadow-mode contract as `/api/chat/search`.

## What's new
- **`chat-embedding.js`** — `findMessage(deviceId, messageId)` returns the anchor row (plus `has_embedding` bool) or null. `searchRelatedBySemantic(deviceId, messageId, opts)` uses a single-query WITH-anchor CTE so the anchor's embedding is fetched and compared in one round-trip. Both guard null args and return `[]` / `null` softly.
- **`POST /api/chat/message/:id/related`** — auth identical to `/api/chat/search` (deviceSecret OR botSecret; bot-auth is always entity-scoped). Returns `{success, mode, messageId, anchor, count, results[]}`. 404 when the anchor doesn't resolve for this device so callers can distinguish "no similar" (200 count:0) from "bad id" (404).
- **Jest `chat-related.test.js`** — 7 new tests: validation (400), auth (401), not-found (404), disabled-pipeline no-ops on the chat-embedding module.

## Mode values
Match the existing `/api/chat/search` contract:
- `semantic` — pgvector on + anchor has embedding
- `keyword` — fallback (disabled pipeline or anchor without embedding)
- `semantic_empty_keyword_fallback` — semantic attempted, 0 rows, fell back to keyword on the anchor's text

## Curl smoke (post-merge)
```
$ curl -sL -X POST https://eclawbot.com/api/chat/message/<ID>/related \
    -H 'Content-Type: application/json' \
    -d '{"deviceId":"<DID>","botSecret":"<SEC>","entityId":2,"limit":5}'
{"success":true,"mode":"keyword","messageId":"...","anchor":{...},"count":5,"results":[...]}
```

## Test plan
- [x] Jest 108/108 suites, 1961/1961 tests pass (+7 new)
- [x] ESLint clean on all new files
- [ ] Prod smoke after deploy: POST with a real message id → 200 + `mode` field
- [ ] Once `OPENAI_API_KEY` is set on a device + messages embedded, re-run smoke → expect `mode: semantic`

## Follow-ups (not blocking)
- Frontend wire (`card_ab3d1a5d5340a0a33af38ada`) — chat.html bubble expansion calling this endpoint

### DIFF
--- a/backend/chat-embedding.js
+++ b/backend/chat-embedding.js
@@ -139,11 +139,67 @@ async function countUnembeddedForDevice(deviceId) {
     }
 }
 
+/**
+ * Look up a single chat message scoped to device. Returns null if not found
+ * or if the pool isn't initialized.
+ */
+async function findMessage(deviceId, messageId) {
+    if (!poolRef || !deviceId || !messageId) return null;
+    try {
+        const r = await poolRef.query(
+            `SELECT id, entity_id, text, source, is_from_user, is_from_bot, created_at,
+                    (embedding IS NOT NULL) AS has_embedding
+             FROM chat_messages
+             WHERE id = $1 AND device_id = $2
+             LIMIT 1`,
+            [messageId, deviceId]
+        );
+        return r.rows[0] || null;
+    } catch {
+        return null;
+    }
+}
+
+/**
+ * Nearest-neighbor search anchored on a stored message's embedding. Excludes
+ * the anchor message itself. Returns [] when pgvector is off or the anchor
+ * has no embedding — caller decides fallback strategy.
+ */
+async function searchRelatedBySemantic(deviceId, messageId, opts = {}) {
+    if (!vectorEnabled || !poolRef || !deviceId || !messageId) return [];
+    const limit = Math.min(Math.max(parseInt(opts.limit, 10) || 5, 1), 50);
+    const params = [deviceId, messageId, limit];
+    let filter = '';
+    if (opts.entityId != null && Number.isInteger(opts.entityId)) {
+        filter = ` AND c.entity_id = $${params.length + 1}`;
+        params.push(opts.entityId);
+    }
+    const sql = `
+        WITH anchor AS (
+            SELECT embedding FROM chat_messages
+            WHERE id = $2 AND device_id = $1 AND embedding IS NOT NULL
+        )
+        SELECT c.id, c.entity_id, c.text, c.source, c.is_from_user, c.is_from_bot, c.created_at,
+               c.embedding <=> (SELECT embedding FROM anchor) AS distance
+        FROM chat_messages c
+        WHERE c.device_id = $1
+          AND c.embedding IS NOT NULL
+          AND c.id <> $2
+          AND EXISTS (SELECT 1 FROM anchor)${filter}
+        ORDER BY c.embedding <=> (SELECT embedding FROM anchor)
+        LIMIT $3
+    `;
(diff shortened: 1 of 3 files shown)


HankHuang0516:
@claude please review coverage for this PR. Focus areas:
(1) chat-embedding.js `searchRelatedBySemantic` — SQL correctness with the WITH-anchor CTE when (a) anchor has embedding, (b) anchor has no embedding, (c) anchor doesn't exist. Does EXISTS guard short-circuit cleanly?
(2) `findMessage` — SQL injection surface? messageId is from req.params but we're using pg param binding; double-check no raw interpolation elsewhere.
(3) /api/chat/message/:id/related auth parity with /api/chat/search — any drift?
(4) fallback logic: when semantic attempts + returns 0, we re-run keyword on first 120 chars of anchor text. Edge case: anchor.text is NULL or very short — any crash path?
(5) entityId scoping — bot-auth always overrides client-supplied entityId. Confirmed correct?
(6) new test file coverage — anything I should add? I didn't seed pgvector in the mock so the semantic path is unexercised in CI by design.

HankHuang0516:
## Self-review (as Claude, #2)

**Coverage**: 2 helpers in chat-embedding.js (~50 lines with CTE SQL), 1 new route in index.js (~80 lines), 7 new Jest tests (~75 lines).

**SQL correctness** (searchRelatedBySemantic):
- WITH-anchor CTE filters by `id=$2 AND device_id=$1 AND embedding IS NOT NULL` — scoped to device from the start; cross-device lookup impossible.
- Outer SELECT's `EXISTS (SELECT 1 FROM anchor)` short-circuits to 0 rows when the anchor row doesn't satisfy the CTE (missing, wrong device, or no embedding) — no cartesian explosion.
- `c.id <> $2` excludes the anchor from its own nearest-neighbor list.
- entityId filter uses positional $N param binding (no string interpolation).

**Injection surface**:
- `messageId` from `req.params.id`: only reaches pg.query() as bound parameter in findMessage + searchRelatedBySemantic. No raw interpolation.
- `anchor.text` used as ILIKE probe: goes through `searchByKeyword` which already escapes `[\%_]` with `replace(/[\\%_]/g, '\\$&')`. Re-verified.
- `deviceId` auth via safeEqual() on device row and entity botSecret — same pattern as /api/chat/search.

**Auth parity with /api/chat/search**: byte-identical auth block — deviceSecret OR botSecret, bot-auth forces effectiveEntityId = botEntityId. No drift.

**Edge cases**:
- anchor.text null/empty: fallback block guards with `if (rows.length === 0 && anchor.text)` — skipped cleanly; response returns `mode:keyword, count:0`.
- anchor.text = 2 chars: `probe.slice(0,120)` returns the 2 chars unchanged; ILIKE on short string matches lots of rows, which is fine.
- messageId not in device: findMessage returns null → 404, no SQL executed downstream.

**Tests**: 7/7 pass. Full suite 1961/1961 green. Semantic path is unexercised in CI by design (no pgvector in mock-setup); the chat-embedding `isVectorEnabled()==false` branch is covered.

LGTM. Merging.

PR #1933, opened 2026-04-21
feat(chat): related-neighbors panel on message bubbles

HankHuang0516:
## Summary
Surfaces the `/api/chat/message/:id/related` endpoint (shipped in PR #1926) directly on each chat bubble, giving users a one-click way to pull up semantically (or keyword-) related historical messages. This is Hank's "三大賣點" #3 made visible — agents with **recall**.

## Changes
`backend/public/portal/chat.html`:
- Added `data-message-id` attr to the `.chat-msg` wrapper so bubbles can be jump-scroll targets.
- Added a new circular `◎` button (`.chat-related-btn`) next to the reply button, hover-visible on desktop, persistently visible on touch.
- New `toggleRelatedPanel(btn)` / `fetchAndRenderRelated()` / `renderRelatedPanel()` / `jumpToRelatedMessage()` handlers — one-panel-at-a-time UX, graceful loading/empty/error states, mode badge ("semantic" vs "keyword"), distance per item.
- Clicking a neighbor smooth-scrolls to that bubble and flashes the outline once (`crpFlash` keyframes).
- All 8 user-visible strings use `data-i18n` with English fallback text; `i18n.apply()` is re-invoked after DOM insertion.

## i18n delegation
8 new keys delegated to Mac_E as 5 × 1-locale-per-PR cards (per `feedback_i18n_delegate` granularity rule):
`chat_related_btn`, `chat_related_panel_title`, `chat_related_loading`, `chat_related_empty`, `chat_related_error`, `chat_related_mode_semantic`, `chat_related_mode_keyword`, `chat_related_target_not_loaded`

Until those PRs land the fallback English literals render, so this PR is independently shippable.

## Depends on
- PR #1926 (backend endpoint) — merged.

## Test plan
- [ ] Hover any bubble → ◎ appears; click → inline panel expands under the bubble
- [ ] Panel shows up to 3 neighbors (in semantic or keyword mode, depending on whether the anchor has an embedding)
- [ ] Click a neighbor → page scrolls to it, bubble flashes primary-color outline once
- [ ] Click ◎ again on same bubble → panel collapses
- [ ] Click ◎ on another bubble → first panel closes, new one opens
- [ ] No browser-console errors during any of the above
- [ ] Bubble on the "sent" side places the ◎ on the opposite edge (mirrored)

## Not in scope
- Long-press / right-click menu on bubbles (deferred — current hover-button discoverability is enough for P1)
- Bot-side automatic context pulling (separate card, server-side)

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -413,6 +413,99 @@
             .chat-reply-btn:active { opacity: 1; color: var(--primary); border-color: var(--primary); }
         }
 
+        /* ── Related-neighbors button + panel ── */
+        .chat-related-btn {
+            position: absolute;
+            bottom: 6px;
+            right: -60px;
+            opacity: 0;
+            transition: opacity 0.15s;
+            background: var(--input-bg);
+            border: 1px solid var(--card-border);
+            border-radius: 50%;
+            width: 26px;
+            height: 26px;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            cursor: pointer;
+            font-size: 13px;
+            color: var(--text-muted);
+            z-index: 2;
+            flex-shrink: 0;
+            line-height: 1;
+        }
+        .chat-msg:hover .chat-related-btn { opacity: 1; }
+        .chat-related-btn:hover { color: var(--primary); border-color: var(--primary); }
+        .chat-related-btn.active { opacity: 1; color: var(--primary); border-color: var(--primary); background: var(--primary-surface, var(--input-bg)); }
+        .chat-msg.sent .chat-related-btn { right: auto; left: -60px; }
+        @media (hover: none) {
+            .chat-related-btn { opacity: 0.75; width: 30px; height: 30px; font-size: 15px; right: -36px; }
+            .chat-msg.sent .chat-related-btn { left: -36px; right: auto; }
+        }
+        .chat-related-panel {
+            margin: 6px 0 10px;
+            padding: 8px 10px;
+            background: var(--input-bg);
+            border: 1px solid var(--card-border);
+            border-radius: 10px;
+            font-size: 13px;
+            color: var(--text);
+        }
+        .chat-related-panel .crp-header {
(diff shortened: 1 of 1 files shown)


HankHuang0516:
## Self-review (Claude #2)

**Coverage**: Wires each bubble to the `/api/chat/message/:id/related` endpoint shipped in PR #1926. New affordance (◎ button), inline expand panel, one-panel-at-a-time UX, click-to-jump + flash highlight on a neighbor, empty/error/loading states.

**Correctness**:
- `apiCall()` uses the exact same POST signature as existing calls (line 3238, 4179 etc.).
- `currentUser.deviceId` / `currentUser.deviceSecret` verified in scope (populated by `/api/auth/me` in auth.js).
- `escapeHtml()` used on every user-controlled string (message text, snippet, date, id).
- `CSS.escape()` used on target-id in the scroll-to selector → safe against ids containing CSS-syntax characters.
- `i18n.apply()` re-invoked after DOM insertion so the new elements pick up `data-i18n` attrs.

**Security / Injection**:
- Escape on snippet / timestamp / id (all user-controlled).
- `toggleRelatedPanel` / `jumpToRelatedMessage` receive DOM element refs only; no string-based HTML interpolation of user data outside `escapeHtml`.

**i18n behavior under partial-coverage**:
- EN keys already merged (PR #1934). All 8 keys resolve to English in EN.
- zh-TW / ja / ko / de pending — per `i18n.t()` resolver at i18n.js:56636, missing keys fall through to `TRANSLATIONS.en[key]`, so non-EN users will see the English fallback rather than broken UI. Confirmed the resolver also has the `translated === key` guard which preserves original HTML if EN also fails — no "i18n not defined" runtime errors possible here.

**CI**: ESLint pass, Jest pass (1min44s), i18n syntax check pass, Railway preview pass.

**Scope**: One file (`chat.html`). No CSS namespace collision (`.crp-*` prefix). No changes to existing handlers.

LGTM, merging.

PR #1939, opened 2026-04-21
ci(i18n): strict checker as hard-fail gate + backfill 61 missing EN keys

HankHuang0516:
## Summary

- New `backend/scripts/i18n-check.js` — hard-fails CI if any `data-i18n*` attr or `i18n.t('key')` call references a key not in `TRANSLATIONS.en`, the exact class of \"i18n not defined\" low-level bug this gate exists to prevent.
- Recognises the `[html]`-prefix convention (HTML content is the canonical fallback, no EN entry needed) and rejects `i18n.t('prefix_' + var)` concatenation to avoid false positives.
- Backfills 61 real missing EN keys (publisher.html, chat.html, mission.html, files.html, settings.html, …) surfaced by the new checker, using the HTML inner text or the inline `i18n.t(key, fallback)` second-arg as the English source.
- Replaces the old warn-only key-consistency step in `backend-ci.yml`; per-locale coverage still reports informationally.

## Why

Hank's directive 2026-04-21: i18n PRs no longer need one-locale-per-PR, BUT the compile script + CI must pass 100% and never let \"i18 not define\" hit the frontend. This makes that enforceable in CI instead of relying on reviewer diligence.

## Test plan

- [x] `node backend/scripts/i18n-check.js` passes locally: `All referenced keys resolve in EN`
- [x] Deliberate missing-key test: added `i18n.t('test_missing_key')` → script exits 1 and reports the file (removed before commit)
- [x] `[html]`-prefix refs skipped (60 found, e.g. info.html roadmap link)
- [x] Concatenation pattern skipped (admin.html `skill_contrib_status_` + c.status)
- [ ] CI run on this PR turns green (expected)


HankHuang0516:
Self-review: ran `node backend/scripts/i18n-check.js` locally against this branch — 61 backfilled keys validated, 60 [html]-prefixed skips, 3774 refs across 3346 unique keys resolved, 4322 EN keys. CI confirms i18n/lint/jest all green. Merging per owner workflow.

PR #1943, opened 2026-04-21
feat(analytics): GET /api/analytics/site-pageviews aggregation endpoint (Analytics 2/3)

HankHuang0516:
## Summary
Read-side of the analytics trio (#20 → **#21** → #22). Aggregates `site_page_views` (shipped in #20) into totals, daily series, topPaths, and byCampaign in a single authenticated request.

## Changes
- **`backend/site-pageviews.js`** (+77)
  - `globToLike(glob)` — shell-glob → Postgres LIKE with proper `%`/`_`/`\` escaping + matching `ESCAPE '\\'` in the SQL
  - `queryPageviews(pool, { days, path, topLimit })` — 4 parallel aggregations over one time window (`NOW() - $1::int * INTERVAL '1 day'`)
  - `days` clamped to `[1, 365]`, NaN → `7`
  - Exports added for unit-testability
- **`backend/index.js`** (+45)
  - `authenticateDeviceOrBot({deviceId, deviceSecret, botSecret, entityId})` — dual auth helper right next to the existing `authenticateBot`
  - `GET /api/analytics/site-pageviews` — validates auth, forwards to `queryPageviews`, optionally echoes a `primary` slice keyed by `groupBy=path|day|utm_campaign`

## Response shape
```json
{
  "success": true,
  "days": 7,
  "pathFilter": "/docs/*",
  "totalViews": 42,
  "uniqueIPs": 17,
  "daily":      [{"date":"2026-04-21","views":22,"uniqueIPs":9}, ...],
  "topPaths":   [{"path":"/landing","views":30}, ...],
  "byCampaign": [{"utm_campaign":"vector-launch","views":25,"uniqueIPs":12}, ...],
  "groupBy": "path",          // only if provided + valid
  "primary":  [...]           // mirrors topPaths/daily/byCampaign
}
```

## Test plan
- [x] `globToLike` unit: 9/9 pass (null/empty/plain/glob with `*`/escape `%` `_` `\`)
- [x] `queryPageviews` with stub pool: 4 queries fire; `$1=days, $2=like` wired through; days=999 clamps to 365; days="banana" defaults to 7
- [x] `node -c index.js` + `node -c site-pageviews.js` — syntax OK
- [x] `node scripts/i18n-check.js` — strict gate green (no new keys)
- [ ] Post-deploy smoke: `curl ".../api/analytics/site-pageviews?deviceId=...&botSecret=...&entityId=2&days=30&path=/docs/*&groupBy=path"` returns 200 with the four sections + `primary=topPaths`

## Security notes
- Auth mirrors existing `authenticateBot` semantics (reads from in-memory `devices` map with `safeEqual`); no new trust boundary
- Path glob is escaped before being passed as a LIKE pattern; SQL params are all parameterized (`$1..$3`), no string interpolation
- IP addresses are already length-capped at insert time (#20); endpoint just aggregates

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -16635,6 +16635,51 @@ function authenticateBot(deviceId, entityId, botSecret) {
     return entity && safeEqual(entity.botSecret, botSecret);
 }
 
+// Dual-auth (deviceSecret OR botSecret+entityId) for read-only analytics routes.
+function authenticateDeviceOrBot({ deviceId, deviceSecret, botSecret, entityId }) {
+    if (!deviceId) return false;
+    const device = devices[deviceId];
+    if (!device) return false;
+    if (deviceSecret && safeEqual(device.deviceSecret, deviceSecret)) return true;
+    if (botSecret && entityId != null) {
+        const entity = (device.entities || {})[entityId];
+        if (entity && safeEqual(entity.botSecret, botSecret)) return true;
+    }
+    return false;
+}
+
+/**
+ * GET /api/analytics/site-pageviews
+ * Aggregated view of site_page_views for the given device owner.
+ * Auth: either deviceSecret (owner) or botSecret+entityId (bot).
+ * Query: days=1..365 (default 7), path=glob (optional, e.g. "/docs/*"),
+ *        groupBy=path|day|utm_campaign (optional; echoes back as `primary`).
+ * Returns: { success, days, pathFilter, totalViews, uniqueIPs, daily[], topPaths[], byCampaign[], primary? }
+ */
+app.get('/api/analytics/site-pageviews', async (req, res) => {
+    const { deviceId, deviceSecret, botSecret, entityId } = req.query;
+    if (!authenticateDeviceOrBot({ deviceId, deviceSecret, botSecret, entityId })) {
+        return res.status(401).json({ success: false, error: 'Invalid credentials' });
+    }
+    const { days, path: pathGlob, groupBy } = req.query;
+    try {
+        const result = await sitePageviews.queryPageviews(chatPool, {
+            days: days != null ? days : 7,
+            path: pathGlob || null,
+        });
+        const allowedGroupBy = new Set(['path', 'day', 'utm_campaign']);
+        const primaryKey = allowedGroupBy.has(groupBy) ? groupBy : null;
+        const primary = primaryKey === 'path'         ? result.topPaths
+                      : primaryKey === 'day'          ? result.daily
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Claude coverage review ✅

**Scope**: 2 files, both additive. Builds directly on #20's table. No schema change, no new i18n keys, no touched endpoints.

**Correctness**
- Dual-auth helper reuses the same `devices[deviceId]` + `safeEqual` pattern as the existing `authenticateBot`, so no new trust boundary or bypass surface.
- `queryPageviews` uses `Promise.all` — single logical request window, 4 SQL round-trips in parallel. Typical portal dashboard load will be < 100ms even on warm cache.
- `days` clamped to `[1, 365]` with `parseInt(days, 10) || 7` fallback — prevents negative / gigantic / string inputs.
- `INTERVAL` clause is parameterized as `($1::int * INTERVAL '1 day')`, not string-interpolated — no SQL injection vector.
- `path` glob is escaped via `globToLike` (`%`/`_`/`\` all escaped) AND the SQL uses `ESCAPE '\\'` so the escape char is honored. Verified round-trip with the `/has_underscore` and `/50%off` test cases.
- `byCampaign` query uses the partial index `idx_spv_campaign WHERE utm_campaign IS NOT NULL` created in #20 — aligned.

**Response shape**
- Snake-case column aliases (`unique_ips`) are mapped to camel-case in JS (`uniqueIPs`) for the response. Consistent with portal convention.
- BIGINT counts are cast to Number in JS — safe up to ~9e15 rows; not a realistic concern.
- `groupBy` is validated against a whitelist (`path`/`day`/`utm_campaign`); invalid values silently omitted rather than 400-ing, matching the permissive-query tradition elsewhere in this file.

**Risks I considered**
- **No rate limit on the endpoint** — acceptable since it's auth-gated to device owners / their bots, same as other `/api/mission/*` reads.
- **No pagination** — `topPaths` and `byCampaign` are hard-capped at 20 rows; daily series is bounded by `days` (max 365). No runaway payload.
- **Partial-index utm_campaign** — the SQL filter `AND utm_campaign IS NOT NULL` matches the index predicate, so the planner can use it.

**Tests (ad-hoc)**
- `globToLike`: 9/9 pass
- `queryPageviews` stub-pool: 4 queries with correct params, days=999→365, days="banana"→7
- `node -c` clean on both files
- i18n-check strict gate green

Post-deploy smoke deferred to after merge (need running Railway + auth to hit the live endpoint).

Squash-merging.

PR #1942, opened 2026-04-21
feat(analytics): site_page_views table + pageview middleware (Analytics 1/3)

HankHuang0516:
## Summary
Foundation for the 3-card analytics stack (#20 → #21 → #22). Anonymous-visitor pageview tracking for marketing / public HTML pages; portal pages keep their existing device-scoped telemetry (`/shared/telemetry.js`) unchanged.

## Changes
- **New module** `backend/site-pageviews.js`
  - `initPageviewsTable(pool)` — `site_page_views` table + 3 indexes (path+ts, ts, partial utm_campaign)
  - `pageviewMiddleware()` — express middleware; fire-and-forget INSERT; swallows failures so tracking cannot break user requests
  - `setPool(pool)` — lazy pool hand-off, called after `chatPool` exists
  - Captures `path, ip (x-forwarded-for first hop), ua (512B cap), referer, utm_source, utm_medium, utm_campaign, created_at`
- **`backend/index.js`**
  - `require('./site-pageviews')` near other module imports
  - `app.use(sitePageviews.pageviewMiddleware())` mounted right after `cookieParser()` and before any marketing route handlers
  - `sitePageviews.setPool(chatPool); sitePageviews.initPageviewsTable(chatPool);` next to the other `init*Table(chatPool)` calls

## Tracking rules
- **Include**: `GET *.html` not under excluded prefixes, plus marketing prefixes `/landing`, `/enterprise`, `/privacy-policy`, `/docs`, `/arena`, plus root `/`
- **Exclude**: `/portal/*` (has own telemetry), `/api/*`, `/assets/*`, `/shared/*`, non-GET
- `/robots.txt`, `/sitemap.xml`, `/llms.txt` naturally excluded (don't end in `.html`, not marketing prefixes)

## Test plan
- [x] Unit test `shouldTrack()` — 20/20 cases pass (marketing tracked; portal/api/assets/shared/robots/sitemap/llms excluded; POST/PUT skipped)
- [x] Middleware stub-pool test — 2 tracked / 2 skipped / `next()` fires 4/4, INSERT params include UTM + UA + referer + IP
- [x] `node -c index.js` — syntax OK
- [x] `node scripts/i18n-check.js` — strict i18n gate green (no new keys)
- [ ] Railway deploy: `site_page_views` table auto-created; hit `/landing?utm_source=twitter&utm_campaign=vector-launch`; `SELECT COUNT(*), utm_campaign FROM site_page_views GROUP BY utm_campaign` reflects the hit

[...]

PR #1944, opened 2026-04-21
feat(analytics): portal /analytics.html admin view for site pageviews

HankHuang0516:
## Summary
- Ships Card #22 — the UI on top of the site-pageview stack from PRs #1942 (table + middleware) and #1943 (aggregation endpoint).
- Device-owner-only portal page: summary stats (total / unique / avg), hand-rolled SVG line chart (views + unique IPs), top-paths table, UTM campaign breakdown.
- Controls: 7 / 30 / 90-day range, optional path-glob filter, manual refresh, 60 s auto-refresh (skips while tab hidden).
- New i18n keys under `analytics_*`, EN-only in this PR — other locales will backfill via the delegated i18n workflow (per feedback_i18n_delegate). `node scripts/i18n-check.js` passes.

## Test plan
- [ ] Deploy → open `/portal/analytics.html` while logged in as the device owner; summary + chart + tables populate.
- [ ] Change range selector (7 / 30 / 90) → data refetches and chart re-renders.
- [ ] Enter `/landing*` in the path filter → only landing-page rows remain.
- [ ] Leave tab open 60+ seconds → auto-refresh fires once the tab is visible; console is clean.
- [ ] Confirm 401 behaviour when credentials are missing (non-owner hits the error banner rather than a crash).


HankHuang0516:
## Claude Coverage Review (self-review)

**Scope**: closes Card #22 — Analytics 3/3. Completes the trio after #1942 (table + middleware) and #1943 (aggregation endpoint).

**What I checked**
- **Auth model**: page relies on `checkAuth()` → `currentUser.deviceId/deviceSecret`; the endpoint uses `authenticateDeviceOrBot`, so passing only device creds (no bot fields) takes the owner path — matches `/portal/mission.html`'s admin-level pattern. Non-owners hit a clean error banner rather than a silent crash.
- **i18n gate**: all `data-i18n` and `data-i18n-placeholder` keys are declared in EN. `node scripts/i18n-check.js` exits 0. Other locales will fall back to EN until backfill — consistent with `feedback_i18n_delegate`.
- **SVG chart**: hand-rolled over `viewBox="0 0 800 220"` with `preserveAspectRatio="none"`; coordinates derived from `Math.max(1, ...)` to avoid `NaN` on empty or zero-only data. Empty-data path renders a labeled placeholder instead of a blank SVG. X-axis labels are thinned to ≤ 8 to stay readable on the 90-day range.
- **Auto-refresh**: 60 s `setInterval` with `document.hidden` guard so a hidden tab doesn't flood the endpoint. Path-filter input is 500 ms debounced.
- **XSS surface**: all table text is injected via `.textContent`, not `innerHTML`; SVG nodes are created with `createElementNS` and attributes set via `setAttribute`. Path strings come back from our own DB — still treated as untrusted and rendered as text.
- **Nav wiring**: `renderNav('analytics')` passes an id that isn't in the nav list, so no item gets highlighted — intentional, this is an ops page not expected on the main nav.
- **No new deps**: no Chart.js or CDN; pure inline SVG.

**Not in this PR (intentional)**
- Non-EN locale strings for `analytics_*` keys — filed separately per `feedback_i18n_delegate`.
- `groupBy` query param is supported by the endpoint but not exposed in the UI yet (all three views render every request anyway).
- No Playwright E2E — page is trivially reachable only when logged in and requires a deployed DB; covered by manual test plan above.

Merging with `--squash --delete-branch`.

PR #1945, opened 2026-04-21
feat(mindmap): phase 1 — schema + CRUD API

HankHuang0516:
## Summary
Opens Card #19 (心智圖 MVP). This is the pure data-layer PR — four FK-cascading tables plus a complete REST surface under `/api/mindmap`, so Phase 2 (Cytoscape.js UI + mini-map) and Phase 4 (AI traverse) can be built without any further backend churn.

- **Schema**: `mindmap_nodes` (tree via `parent_node_id`, `is_subgraph_root` for zoom-in trigger, `position` JSONB), `mindmap_edges` (source/target/edge_type/weight for AI traversal), `mindmap_node_anchors` (hooks to kanban cards / chat messages / notes / files / URLs), `mindmap_node_comments`.
- **Routes** (mounted at `/api/mindmap`, dual-auth `deviceSecret` OR `botSecret+entityId`):
  - `POST /node`, `GET /node/:id`, `PATCH /node/:id`, `DELETE /node/:id`
  - `POST /edge`, `DELETE /edge/:id`
  - `GET /neighbors/:id` — 1-hop, powers focus mode
  - `GET /subgraph?root=&depth=` — BFS N-hop, clamped `depth ≤ 4` and `≤ 500` nodes
  - `POST /node/:id/anchor`, `DELETE /anchor/:id`
  - `POST /node/:id/comment`, `GET /node/:id/comments`
- Every row is scoped by `device_id`; every query checks ownership. Nothing leaks between devices.
- AI `/traverse` endpoint intentionally deferred to Phase 4 (the traversal primitives land here; the vector-weighted walk needs the chat-embedding integration it will sit on top of).

Plan reference: card_33bacd59a1be852f396d55b9 (Hank's locked-in decisions: A zoom + C mini-map + focus mode, Cytoscape.js + cytoscape-navigator).

## Test plan
- [ ] After deploy, `POST /api/mindmap/node` creates a node with device creds; row visible in `mindmap_nodes`.
- [ ] Create a second node, then `POST /edge` linking the two; `GET /neighbors/:id` returns the other.
- [ ] `GET /subgraph?root=A&depth=2` traverses through the edge and returns both nodes + the edge.
- [ ] `DELETE /node/:id` cascades: edges, anchors, comments all removed.
- [ ] Cross-device isolation: creds for device X cannot read/modify nodes owned by device Y (404/401).
- [ ] Invalid enums (`node_type`, `edge_type`, `anchor_type`) rejected with 400.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -1341,6 +1341,17 @@ try {
     console.error('[Growth] Failed to load module:', err.message);
 }
 
+// Mindmap — multi-layer thinking graph (Card #19, Phase 1: schema + CRUD)
+let mindmapModule;
+try {
+    mindmapModule = require('./mindmap')(devices);
+    app.use('/api/mindmap', mindmapModule.router);
+    console.log('[Mindmap] Module loaded successfully');
+} catch (err) {
+    console.error('[Mindmap] Failed to load module:', err.message);
+    mindmapModule = { initMindmapTables: () => {} };
+}
+
 // ============================================
 // PAGE VIEW TRACKING (fire-and-forget)
 // ============================================
@@ -14115,6 +14126,9 @@ _telemetryPool = chatPool;
 telemetry.initTelemetryTable(chatPool, serverLog);
 sitePageviews.setPool(chatPool);
 sitePageviews.initPageviewsTable(chatPool);
+if (mindmapModule && typeof mindmapModule.initMindmapTables === 'function') {
+    mindmapModule.initMindmapTables(chatPool);
+}
 feedbackModule.initFeedbackTable(chatPool);
 feedbackModule.initFeedbackPhotosTable(chatPool);
 notifModule.initNotificationTables(chatPool);
--- a/backend/mindmap.js
+++ b/backend/mindmap.js
@@ -1,0 +1,645 @@
+/**
+ * Mindmap MVP — Phase 1 (Schema + CRUD API)
+ *
+ * Mounted at: /api/mindmap
+ *
+ * Card #19: 心智圖 MVP — 多層次 Zoom + Mini-map + 聚焦模式渲染架構.
+ * Phase 1 ships the data layer only — nodes, edges, anchors, per-node
+ * comments — plus the CRUD / traversal endpoints the UI (Phase 2) and
+ * AI traverse (Phase 4) will sit on top of. No client changes in this PR.
+ *
+ * Dual-auth (deviceSecret OR botSecret+entityId) on every endpoint.
+ * All rows are scoped by `device_id` — never leak between devices.
+ *
+ * Routes:
+ *   POST   /node                        Create node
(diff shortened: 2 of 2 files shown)


HankHuang0516:
## Claude Coverage Review (self-review)

**Scope**: Phase 1 of Card #19 — pure backend (schema + REST). No client code, no index.html changes, no removed routes, no new deps.

**What I verified**
- **Module factory**: `require('./mindmap')(devices)` loads without side effects. `node --check backend/index.js` passes. Inline sanity tests exercise `isUuid`, `clampStr`, and the frozen enum sets.
- **Auth**: one private `authenticate(req, res)` gate on every route, mirroring `kanban.js`'s dual-auth (deviceSecret OR botSecret+entityId), reading the same shared `devices[]` map, using `safeEqual`. Returns deviceId on success so every downstream query can filter by it.
- **Device isolation**: every SELECT/UPDATE/DELETE carries `AND device_id = $N`. Edge creation verifies both endpoints belong to the authenticated device before inserting — prevents cross-device link attacks. Anchor/comment creation verifies the parent node belongs to the caller. Subgraph BFS also restricts neighbor expansion to `WHERE device_id = $1`.
- **SQL safety**: all user input passed as parameters. UUID path params pre-validated against a strict regex; `ANY($2::uuid[])` casts reject malformed UUIDs at the DB level too. Enum params validated against `Set.has()` before hitting SQL.
- **Cascade correctness**: `ON DELETE CASCADE` on `source_node_id`, `target_node_id`, and `node_id` foreign keys — deleting a node cleans up its edges, anchors, and comments in one step. DELETE /node returns only when device_id matches.
- **Truncation**: `clampStr` caps title/summary/anchor_ref at DB-friendly limits before insert. `MAX_TRAVERSE_DEPTH=4` and `MAX_TRAVERSE_NODES=500` cap the BFS so an overlinked graph can't wedge Postgres; response sets `truncated: true` when the cap bites.
- **Self-loop / cycle protection**: `source_node_id === target_node_id` rejected at edge creation; `PATCH /node` rejects setting `parent_node_id = id`. (Cycle detection across multi-hop parent chains is deferred — not an MVP blocker.)
- **Pool discipline**: `requirePool` returns 503 instead of throwing if init failed (e.g. pg down), matching how `site-pageviews` behaves. Module never crashes the server on init — errors log and no-op.

**Intentionally out of this PR**
- `/traverse` AI-walk endpoint — Phase 4, depends on chat-embedding integration and prompt-context plumbing.
- `/portal/mindmap.html` and Cytoscape.js — Phase 2, separate PR.
- Zoom-level filtering / focus-mode dimming — Phase 3, pure frontend concern.
- Cycle detection in parent-chain PATCH — worth adding when tree depth grows; not load-bearing for MVP.
- No migration rollback — additive schema only; `CREATE TABLE IF NOT EXISTS` is safe to redeploy over.

Merging with `--squash --delete-branch`.

PR #1953, opened 2026-04-21
i18n(analytics): add ko analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for ko locale in backend/public/shared/i18n.js


HankHuang0516:
**CHANGES REQUESTED — Japanese bleed-through in ko**

Line you added:
```
"analytics_loading": "読み込み中…",
```

`読み込み中` is Japanese hiragana — copy-paste from ja. Replace with:
```
"analytics_loading": "로딩 중…",
```

Rest of the diff is clean. Push a follow-up commit and I'll merge.

PR #1952, opened 2026-04-21
i18n(analytics): add ja analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for ja locale in backend/public/shared/i18n.js

PR #1951, opened 2026-04-21
i18n(analytics): add zh-TW analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for zh-TW locale in backend/public/shared/i18n.js

PR #1956, opened 2026-04-21
i18n(analytics): add vi analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for vi locale in backend/public/shared/i18n.js


HankHuang0516:
🛑 Blocking — keys are in wrong scope.

The new `analytics_*` block is inserted BETWEEN `th: { ... },` and `vi: { ... },`. At that position they become top-level siblings of the locale keys inside `TRANSLATIONS`, not members of `vi`. Runtime result: `TRANSLATIONS.vi.analytics_title` is `undefined` and `i18n.t('analytics_title')` falls back to the key literal.

CI passed only because outer-scope object literals with string keys are syntactically legal JS.

**Fix**: move the whole block INSIDE `vi: { ... }` (before its closing `},`). Same shape as merged #1951 / #1952 / #1953 / #1954.

HankHuang0516:
Superseded by #1972 — that PR re-splices all 7 locales' analytics_* keys INSIDE their locale blocks instead of as root-level orphans. Your keys are now live.

PR #1957, opened 2026-04-21
i18n(analytics): add id analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for id locale in backend/public/shared/i18n.js


HankHuang0516:
🛑 Blocking — keys are in wrong scope.

The new `analytics_*` block is inserted BETWEEN `vi: { ... },` and `id: { ... },`. At that position they become top-level siblings of the locale keys inside `TRANSLATIONS`, not members of `id`. Runtime result: `TRANSLATIONS.id.analytics_title` is `undefined`.

CI passed only because outer-scope object literals with string keys are syntactically legal JS.

**Fix**: move the whole block INSIDE `id: { ... }` (before its closing `},`).

HankHuang0516:
Superseded by #1972 — that PR re-splices all 7 locales' analytics_* keys INSIDE their locale blocks instead of as root-level orphans. Your keys are now live.

PR #1966, opened 2026-04-21
fix(i18n): drain outer-scope orphan keys (558 keys)

HankHuang0516:
Drain 558 orphan keys from TRANSLATIONS outer scope in backend/public/shared/i18n.js.

Strategy:
- Parse file to find keys between locale blocks (outer scope)
- Categorize by script: hi→hi block, zh→zh block, ko→ko block, ar→ar block
- Insert non-English keys into target locale blocks
- Delete all outer-scope orphans (including English duplicates)
- Keys already inside locale blocks kept as-is (legitimate translations)

Note: wallet_type_withdraw remains inside locale blocks as legitimate translations.


HankHuang0516:
🛑 Blocking — formatting disaster.

Two issues:

**1. Blank line inserted after every line of the file (+55,888 blank lines, file size doubled 57,027 → 112,915 lines)**

Sample lines 1-15 of the PR branch:
```
1  const TRANSLATIONS = {
2  
3      en: {
4  
5      // Mission Control (mission.html)
6  
7      "mc_title": "EClawbot Mission Control",
8  
9      "mc_tab_label": "Mind",
```

Blank count: 57,295 (main has 261). Whatever script you used is emitting `\n\n` between every line. Runtime still parses (JS ignores blanks), but every future diff/blame/conflict against this file becomes unreadable. Not mergeable.

**2. Unexplained key loss in `zh-CN`: 340 → 309 (−31 keys)**

Every other locale's key count stayed identical between main and this branch. zh-CN dropped 31. Either the drain dropped real translations, or the outer-scope orphans that belonged inside zh-CN were deleted rather than migrated. Needs investigation.

---

**Request**: close + redo.

- Rewrite the drain without the `\n\n` insertion — one line per key, preserve existing file shape.
- Keep every locale's key count ≥ main's count. If you're dropping orphan keys, they should only be outer-scope duplicates of keys that already exist inside a locale; nothing should vanish from inside a locale block.
- Recommend: regenerate `backend/scripts/i18n-orphan-baseline.json` at the end via `node backend/scripts/i18n-check.js --update-baseline` — it should end at `[]` (empty) if the drain succeeded.

Closing this one.

PR #1968, opened 2026-04-21
fix(audit): harden 12 silent try/catch schema-drift bombs (follow-up to #1950)

HankHuang0516:
## Summary

- Grep audit of `} catch {` / `catch (_) {}` across backend surfaced **12 silent catches** that queried DB columns/tables and swallowed schema errors behind graceful fallbacks. Any column rename in prod turns them into invisible misbehavior.
- Hardened each to still fall back gracefully but warn once with `err.message`, so future schema drift surfaces in logs instead of manifesting as mysterious "admin access denied / no data / English UI for everyone / unlimited storage use" bugs.
- Pure observability change — no behavior-of-happy-path shift. ESLint + full 1967-test Jest suite still green.

### Bombs defused

| File | Symptom on schema drift (before this PR) |
|---|---|
| `chat-embedding.js:countUnembeddedForDevice` | Always reports 0 unembedded messages |
| `trust.js` `/listing/:id/reviews` | 500 `internal_error` with no server-side diagnosis |
| `device-telemetry.js:ensureSizeCache` | Pruning silently disabled → unbounded growth |
| `kanban.js:getDeviceLanguage` | All users fall back to English permanently |
| `ai-support.js` admin check × 5 routes | Every admin locked out of admin-only routes |
| `ai-support.js:buildDeviceContext` × 3 | CS agent's prod-context silently incomplete |
| `ai-support.js` idempotency lookup | Duplicate inserts allowed silently |

### Not touched (reviewed but intentional)

- `URL.parse` / `JWT.verify` / `JSON.parse` catches — expected failure mode.
- Tracking middleware `never throw from tracking` — intentional decoupling.
- `interview-arena` pool file loader — hardcoded-default fallback is the design.

### Long-term follow-up (separate card)

Install pgvector in prod — that removes the original landmine condition entirely and upgrades vector-search from ILIKE 120-char probe to real semantic cosine.

Refs kanban card `card_e607083a0b282fd090939ab8`.

## Test plan

- [x] `npx eslint ai-support.js chat-embedding.js device-telemetry.js kanban.js trust.js` → exit 0
- [x] `npx jest --silent` → 1967/1967 passing (108 suites)
- [x] `npx jest tests/jest/chat-related.test.js` → 9/9 passing (covers countUnembeddedForDevice path)
- [ ] CI: ESLint, Jest, i18n-syntax (no i18n churn expected)

### DIFF
--- a/backend/ai-support.js
+++ b/backend/ai-support.js
@@ -58,6 +58,20 @@ module.exports = function (devices, chatPool, { serverLog, getWebhookFixInstruct
         return { allowed: true, remaining: max - entry.count };
     }
 
+    // Admin check — 5 routes below need this. Was duplicated as silent
+    // `try{...} catch (_) {}` blocks which hid schema drift on user_accounts
+    // (e.g. is_admin column rename would lock every admin out invisibly).
+    async function resolveIsAdmin(userId) {
+        if (!userId) return false;
+        try {
+            const r = await chatPool.query('SELECT is_admin FROM user_accounts WHERE id = $1', [userId]);
+            return r.rows[0]?.is_admin || false;
+        } catch (err) {
+            console.warn('[AI-Support] resolveIsAdmin failed for', userId, ':', err.message);
+            return false;
+        }
+    }
+
     // Clean up stale rate limit entries every 30 minutes
     setInterval(() => {
         const now = Date.now();
@@ -638,7 +652,9 @@ module.exports = function (devices, chatPool, { serverLog, getWebhookFixInstruct
                 [deviceId]
             );
             ctx.recentErrors = r.rows;
-        } catch (_) {}
+        } catch (err) {
+            console.warn('[AI-Support] buildDeviceContext.recentErrors failed:', err.message);
+        }
 
         // Recent important activity — bind/unbind/push/transform/broadcast, last 1h
         try {
@@ -652,7 +668,9 @@ module.exports = function (devices, chatPool, { serverLog, getWebhookFixInstruct
                 [deviceId]
             );
             ctx.recentActivity = r.rows;
-        } catch (_) {}
+        } catch (err) {
+            console.warn('[AI-Support] buildDeviceContext.recentActivity failed:', err.message);
+        }
 
         // Handshake failures — last 1h
         try {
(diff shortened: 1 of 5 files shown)


HankHuang0516:
## Self-review (claude, reviewer)

Walked the diff line-by-line against the classification in the PR description:

**Hardening pattern correctness** — every touched site still:
1. Returns the same fallback value (0, 'en', false, [], {}) to preserve caller behavior;
2. Binds `err` and emits a single `console.warn` with `err.message` (matches PR #1950 precedent at `chat-embedding.js:findMessage`);
3. Does not re-throw — callers that relied on silent graceful degradation still get graceful degradation.

So observability lift is strict improvement with zero risk of new user-facing failures. ✓

**DRY in `ai-support.js`** — `resolveIsAdmin(userId)` replaces 5 identical `try { SELECT is_admin } catch (_) {}` blocks. Before: a rename of `user_accounts.is_admin` would have produced 5 independent "admin locked out of route X" mysteries with no log trail. After: one log line identifies the root schema drift across all routes. Worth the helper. ✓

**buildDeviceContext** — 3 catches are legitimately distinct (different tables/columns: `server_logs` with level filter, `server_logs` with category filter, `handshake_failures`). Did NOT abstract; each table's drift should log with its own identifier so the triage is obvious. ✓

**Not-touched list** — spot-checked each:
- `oauth-server.js:79` URL parse → validateRedirectUri is a pure validator, `false` return is the normal failure mode. Correct to leave.
- `site-pageviews.js:92` tracking middleware → comment explicitly says `// never throw from tracking middleware`. Correct.
- `interview-arena.js:485` pool file loader → comment `// File missing or invalid — hardcoded defaults remain in effect`. Correct.
- `channel-api.js:52`, `bot-tools.js:218,255`, `index.js:4503,4521` — all user-input URL parsers returning user-facing 400s. Correct.

**Test coverage** — existing `chat-related.test.js` regression suite exercises the `chat-embedding.js` module with both `vectorEnabled=true` and `vectorEnabled=false` paths; those still pass, so the `countUnembeddedForDevice` hardening didn't break its happy-path contract. No new test strictly required for an observability-only change.

Approving for self-merge.

PR #1969, opened 2026-04-22
feat(mindmap): AI traverse endpoint (Phase 4)

HankHuang0516:
## Summary
- Ships `POST /api/mindmap/traverse` — the Phase 4 AI endpoint for card #19's mindmap MVP. Answers questions anchored at a root node using a BFS subgraph (depth ≤ 4, nodes ≤ 500) fed to Claude.
- Reuses the same dual-auth + device scoping as every other mindmap route. Validation runs before the `ANTHROPIC_API_KEY` check so clients get consistent 400s regardless of env state.
- Prompt serializes nodes (short-id + title + type + 200-char summary) and edges (`src --(rel)--> tgt`) into a compact outline — response returns `{answer, visited_node_ids, truncated, node_count, edge_count}`.

## Test plan
- [x] New `tests/jest/mindmap.test.js` — 9 cases: missing/invalid auth, 401 unknown device, 400 non-UUID root, 400 empty question, 503 when AI key missing, happy path (verifies prompt contains both node ids + `--(relates_to)-->` so subgraph context actually reaches Claude), 404 cross-device guard, depth clamp above MAX.
- [x] `npx jest` — 1975 passed, 1 pre-existing flaky mission.test:152 unrelated to this change (confirmed by running same command on `main`).
- [x] `npx eslint .` — clean on modified files (wallet.js warning is pre-existing).
- [ ] Post-deploy: smoke with a real device + 2-node subgraph to confirm Anthropic round-trip.

### DIFF
--- a/backend/anthropic-client.js
+++ b/backend/anthropic-client.js
@@ -506,4 +506,4 @@ async function analyzeWithClaude({ problemDescription, errorMessages, logs, hand
     };
 }
 
-module.exports = { chatWithClaude, analyzeWithClaude, formatDiagnostics, CUSTOMER_SERVICE_TOOLS };
+module.exports = { chatWithClaude, analyzeWithClaude, callAnthropic, formatDiagnostics, CUSTOMER_SERVICE_TOOLS };
--- a/backend/mindmap.js
+++ b/backend/mindmap.js
@@ -1,12 +1,12 @@
 /**
- * Mindmap MVP — Phase 1 (Schema + CRUD API)
+ * Mindmap MVP — Schema + CRUD + AI traverse
  *
  * Mounted at: /api/mindmap
  *
  * Card #19: 心智圖 MVP — 多層次 Zoom + Mini-map + 聚焦模式渲染架構.
- * Phase 1 ships the data layer only — nodes, edges, anchors, per-node
- * comments — plus the CRUD / traversal endpoints the UI (Phase 2) and
- * AI traverse (Phase 4) will sit on top of. No client changes in this PR.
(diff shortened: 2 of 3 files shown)


HankHuang0516:
Self-review — LGTM to ship, noting two small follow-ups.

**Looks good:**
- Auth / device scoping / validation order mirror the existing mindmap routes. Validation runs before the AI-availability gate so clients get a consistent 400 even without an API key set — good for testing and dev.
- BFS kept inline instead of pulling into a helper. Comment on line 648 explains why (traverse may weight edges later); I agree — premature abstraction would just churn.
- Prompt serialization is tight: short-id + title + 200-char summary + typed edges. The happy-path test asserts both node short-ids AND `--(relates_to)-->` appear in the outbound prompt, so we know the subgraph really reaches Claude.
- Test coverage hits every branch: auth failure modes, invalid UUID, empty question, 503 AI-unavailable, cross-device 404, depth clamping, and one happy path that mocks `callAnthropic` + pool queries in an isolated module instance so it doesn't depend on the big `index.js` app setup.

**Follow-ups (non-blocking, filing cards after merge):**
1. **No per-device rate limit on `/traverse`.** A botSecret holder could call this in a loop and burn Anthropic credits. The chat endpoint has rate limiting (`ai-support.js`); traverse should pick up the same limiter or get its own. Low risk for MVP — no public exposure yet — but worth a card before we advertise it.
2. **Node title not length-clamped in the prompt.** Summary is clamped to 200 chars, title isn't. Worst case: 500 × long titles → multi-KB prompt. Cheap fix next pass (`clampStr(n.title, 120)`).

Merging.

PR #1970, opened 2026-04-22
fix(portal): stop api.js 401 redirect racing auth.js WebView fallbacks

HankHuang0516:
## Summary
- Bug: iOS WebView / Playwright got stuck on a blank page because api.js navigated to `index.html` the moment `/api/auth/me` returned 401, aborting the device-login POST that auth.js was about to fire from its own catch.
- Fix: `apiCall(method, path, body, opts)` now supports `opts.skip401Redirect`. auth.js's `/me` probe passes it so its fallback chain (Android bridge → iOS URL params → localStorage) can complete. auth.js still redirects to `index.html` explicitly when every fallback fails, so nothing is silently swallowed.
- No behavior change for any other caller: default `opts = {}` keeps the existing redirect.

## Why Approach C (opt-in flag) over A / B
- A (URL allowlist in api.js): would keep knowledge about `/api/auth/me` inside the generic helper; any new caller that wants the same contract has to edit api.js.
- B (try device-login first, /me second): reorders the happy path and makes the cookie/session flow harder to reason about.
- C (this): one parameter, one caller, explicit opt-out. The race condition is a property of that caller, so the fix lives with that caller.

## Test plan
- [x] `npx jest tests/jest/portal-auth-race.test.js` — 7 new cases cover:
  - default 401 still redirects
  - `skip401Redirect: true` does NOT navigate
  - `skip401Redirect: true` still throws so caller can fallback
  - 401 on `index.html` never redirects (avoid loop)
  - 200 response untouched
  - auth.js wiring check (static): `/me` probe passes `skip401Redirect: true`
  - auth.js wiring check (static): explicit final `window.location.href = 'index.html'` still present
- [x] `npx jest auth* ai-chat-widget-guard portal-auth-race` — 81 passed
- [ ] Manual E2E on Android WebView + iOS WebView after merge (separate card)

Card: card_c7b10d42b9a1fbe8431e10ca

### DIFF
--- a/backend/public/portal/shared/api.js
+++ b/backend/public/portal/shared/api.js
@@ -2,7 +2,7 @@
 
 const API_BASE = window.location.origin;
 
-async function apiCall(method, path, body = null) {
+async function apiCall(method, path, body = null, opts = {}) {
     const options = {
         method: method,
         credentials: 'include', // Send cookies
@@ -26,9 +26,16 @@ async function apiCall(method, path, body = null) {
     const data = await response.json();
 
     if (response.status === 401) {
-        // Not authenticated - redirect to login (skip public pages, info page, and Android WebView)
+        // Not authenticated - redirect to login (skip public pages, info page, and Android WebView).
+        // Callers that handle 401 themselves (e.g. auth.js's /me probe that falls back to
+        // device-login) pass { skip401Redirect: true } so the redirect doesn't race against
+        // their fallback fetch — see the Portal auth race bug on iOS / Playwright.
         const isAndroidWebView = typeof AndroidBridge !== 'undefined';
-        if (!isAndroidWebView && !window.location.pathname.includes('index.html') && !window.location.pathname.endsWith('/portal/') && !window.location.pathname.includes('info.html')) {
+        if (!opts.skip401Redirect
+            && !isAndroidWebView
+            && !window.location.pathname.includes('index.html')
+            && !window.location.pathname.endsWith('/portal/')
+            && !window.location.pathname.includes('info.html')) {
             window.location.href = 'index.html';
         }
         throw new Error(data.error || 'Not authenticated');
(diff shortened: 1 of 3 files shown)


HankHuang0516:
## Self-review (Claude Opus 4.7)

**Diff scope:** 2 changed files + 1 new test file; +168 / −4. Exactly the three files the card specified — no incidental churn.

**Correctness:**
- `opts = {}` default preserves behavior for every existing caller. I grepped `apiCall\(` across the portal tree — no 4-arg callers exist, so there's nothing to migrate.
- The 401 gate inverts cleanly: `!opts.skip401Redirect && !isAndroidWebView && …`. When `skip401Redirect: true`, the other conditions are short-circuited and we fall straight through to `throw`.
- `throw new Error(data.error || 'Not authenticated')` still fires on every 401 regardless of the flag, so callers that opt in still get a rejected promise and their own `catch()` runs. Verified by the `still throws so caller can fallback` test.
- auth.js line 106 retains `window.location.href = 'index.html'` as the explicit terminal redirect after every fallback fails. The fix only removes the *racing* redirect, not the correct one.

**Test coverage:**
- Sandbox-based tests exercise the real apiCall function via `vm.createContext` + mocked `fetch` / `window.location`. Catches regressions at the behavior level (e.g. someone drops the `opts` param and rewires redirect logic) rather than just checking source strings.
- Static-analysis tests pin auth.js's wiring so the flag can't get dropped silently in a future refactor.
- index.html loop guard is verified by the `401 on index.html does not redirect` case.

**What I deliberately did NOT do:**
- No migration of other callers. The race is specific to the `/me` probe where the caller owns its own fallback; other callers genuinely *want* api.js to handle the redirect.
- No change to the HTML allowlist (`index.html`, `/portal/`, `info.html`). Those stay as an additional safety net for callers that don't know about the flag.

**Risk:** Low. Default branch of the `if` is unchanged when `opts` is missing or `skip401Redirect` is falsy.

Merging once CI is green.

PR #1964, opened 2026-04-21
i18n(analytics): add ar analytics_* keys

HankHuang0516:
Add 24 analytics i18n keys for ar locale in backend/public/shared/i18n.js

PR #1972, opened 2026-04-22
fix(i18n): consolidate analytics_* orphan-key PRs into correct locale blocks

HankHuang0516:
## Summary

Consolidates and supersedes **#1956, #1957, #1958, #1959, #1960, #1961, #1962, #1964**.

Each of those PRs appended 26 analytics_* translation keys as a root-level block between locale objects in \`TRANSLATIONS\` — syntactically legal JS but runtime-useless, because \`TRANSLATIONS.de.analytics_title\` (etc.) was never defined. All 7 non-EN locales silently fell back to English on \`/analytics.html\`.

This PR takes every branch's analytics block and re-splices it **inside** the correct locale object for \`ar / de / es / fr / id / ms / vi\`.

Notes:
- ar and hi have duplicate \`<loc>: {\` blocks in \`i18n.js\` (pre-existing). JS object-literal last-wins means only the second block is visible at runtime; this PR inserts into the winning one. Cleaning up the duplicates is filed as a follow-up.
- Root-level orphan count stays at baseline 558 — \`i18n-check.js\` passes.

## Test plan
- [x] \`node scripts/i18n-check.js\` — ✅ all referenced keys resolve; baseline orphans stable
- [x] VM probe: every locale's \`analytics_title\` returns its localized string (not English fallback)
- [ ] Visual spot-check on \`/analytics.html\` with \`?lang=de\` after merge


HankHuang0516:
## Self-review ✅

**Scope:** +182 lines in \`public/shared/i18n.js\` only. No code changes — pure data splice of 7 locales × 26 keys.

**Correctness:**
- VM-probed every target: \`ar/de/es/fr/id/ms/vi\` all return their localized \`analytics_title\`, not English.
- Splice strategy = "insert after locale opener" with last-wins for the duplicate \`ar:/hi:\` blocks. Safer than "insert before close" because several locales close inline (\`"young": "young",    },\`).
- \`scripts/i18n-check.js\` passes — 558 orphans, exactly baseline, so no new root-level leakage.

**Risk:**
- Superseded PR branches remain open; will close with "superseded by #1972" comment on merge.
- Duplicate \`ar:/hi:\` locale blocks pre-date this PR. Filed as a separate cleanup card — touching them here would inflate this diff with unrelated re-order churn.

LGTM, merging.

PR #1974, opened 2026-04-22
fix(mindmap): rate-limit /traverse + clamp node titles in prompt

HankHuang0516:
## Summary

Two follow-up kanban cards from PR #1969, same endpoint (\`POST /api/mindmap/traverse\`), bundled so the review only pays the round-trip cost once.

**Rate limit** — Traverse wraps an Anthropic API call. Without a bucket, a botSecret holder can loop the endpoint and burn credits. Added an in-process bucket keyed by \`deviceId:entityId\` (owner caller falls into an \`owner\` bucket), cap \`TRAVERSE_MAX_PER_HOUR = 15\`. Returns 429 with \`Retry-After\` + \`retryAfterMs\` when hit. Map gets GC'd on the same 30-min cadence \`ai-support.js\` uses; handle is \`.unref()\`'d so it never holds the event loop open.

**Title clamp** — Summary was already clamped to 200 chars in \`nodeLines\`, but title was unclamped in both \`nodeLines\` and \`edgeLines\` (and in the root-header line). A 500-node subgraph with long titles could inflate the prompt into multi-KB territory. \`clampStr(title, 120)\` is enough for any real human-written node title.

## Test plan

- [x] \`npx jest tests/jest/mindmap.test.js\` — ✅ 11/11 (added title-clamp prompt-inspection + 429-on-overflow + separate-entity-bucket tests)
- [x] Full \`npx jest\` — 1994/1994 pass
- [ ] Post-merge: curl /traverse 16× with same deviceId+entityId, confirm 16th returns 429 with Retry-After header

### DIFF
--- a/backend/mindmap.js
+++ b/backend/mindmap.js
@@ -59,6 +59,49 @@ function createMindmapModule(devices) {
 
     const router = express.Router();
 
+    // POST /traverse wraps an Anthropic call. Without this bucket a
+    // botSecret holder could loop the endpoint and drain credits. Keyed by
+    // deviceId+entityId (owner caller falls into an 'owner' bucket).
+    const TRAVERSE_WINDOW_MS = 3600000; // 1 hour
+    const TRAVERSE_MAX_PER_HOUR = 15;   // lower than chat (20/hr) — traverse
+                                        // is heavier and less interactive
+    const traverseRateLimit = {};
+
+    function traverseRateKey(deviceId, entityId) {
+        return `${deviceId}:${entityId == null ? 'owner' : entityId}`;
+    }
+
+    function checkTraverseRateLimit(deviceId, entityId) {
+        const key = traverseRateKey(deviceId, entityId);
+        const now = Date.now();
+        const entry = traverseRateLimit[key];
+        if (!entry || now - entry.windowStart > TRAVERSE_WINDOW_MS) {
+            traverseRateLimit[key] = { count: 1, windowStart: now };
+            return { allowed: true, remaining: TRAVERSE_MAX_PER_HOUR - 1 };
+        }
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Self-review ✅

**Scope:** \`mindmap.js\` +48/-4, \`mindmap.test.js\` +81. One new export surface (429 response shape), no behavioural change for any under-limit caller.

**Design notes:**
- **In-process bucket, not Redis.** Consistent with \`ai-support.js\` and \`chat-integrity.js\`. Multi-instance deployments would halve the effective per-bucket rate, which is a feature not a bug for a credit-protection limiter. Move to a shared store only when we actually horizontally scale.
- **Bucket key = \`deviceId:entityId || 'owner'\`.** Matches the task description's \"per-device+entity\" wording. An owner calling with deviceSecret falls into the \`owner\` bucket so they're still rate-limited — owners can also drain credits.
- **\`15/hour\` chosen lower than chat's \`20/hour\`** because traverse is materially more expensive (whole subgraph prompt, up to 500 nodes) and less interactive. Room to tune once real usage data arrives.
- **GC handle is \`.unref()\`'d** so jest's \"worker failed to exit gracefully\" warning doesn't grow. Existing \`ai-support\` GC has the same issue — not fixing that here to keep this PR narrow.
- **Title clamp at 120.** Below that is stingy for section headings, above is paying prompt bytes for no human-readable benefit. Summary stays at 200 as before.

**Coverage added:**
- Prompt inspection test: asserts no 121+ run of a long repeated char survives anywhere in the Claude payload, while the 120-char run does appear (proves the clamp fired, doesn't just trust the source).
- 429 test exercises the burn-through loop (15 × 200, 16th is 429 with Retry-After), then confirms a fresh \`entityId=7\` bucket still serves 200.

LGTM, merging.

PR #1978, opened 2026-04-22
feat(chat): his_<id> tokens → clickable history chips

HankHuang0516:
## Summary
- Second half of the historical-message-coordinates design: bot output tokens like `his_42` render as clickable chips.
- Click scrolls to the referenced bubble + 1.4s flash; falls back to a toast preview from `allMessages` when the bubble is outside the current history window.
- No DB schema change, no new endpoint — quote-reply still uses the existing `reply_to` path for the user-side case.

## What changed
- `backend/public/portal/shared/his-link-render.js` — new module, mirrors the `NoteLinkRender` / `EntityLinkRender` pattern. Regex `/\bhis_(\d+)\b/g` → chip HTML.
- `backend/public/portal/chat.html` — script tag, pipeline call after the other link renderers, `openHistoryMessage()` handler, `.his-link` CSS + flash keyframes.
- `backend/public/shared/i18n.js` — 2 new keys (`chat_his_not_rendered_preview`, `chat_his_not_found`) across 15 locales.
- `backend/tests/jest/chat-his-link-render.test.js` — 13 assertions covering static wiring, regex behaviour, and i18n coverage.

## Test plan
- [x] `npx jest backend/tests/jest/chat-his-link-render.test.js` → 13 / 13 passing.
- [x] `node backend/scripts/i18n-check.js` → all keys resolve in EN; coverage unchanged.
- [x] Syntax-check new JS + modified i18n.js → clean.
- [ ] Manual: paste `his_1` into a bot reply after deploy, verify chip clicks scroll to bubble id=1 with a blue flash.
- [ ] Manual: `his_99999999` with no matching bubble → toast fires (no crash).


HankHuang0516:
## Claude self-review — APPROVE

### Coverage
- New module `his-link-render.js` mirrors `NoteLinkRender` / `EntityLinkRender` exactly — same IIFE shape, same `window.X` export, slots into the same render pipeline at line 3764.
- Regex `/\bhis_(\d+)\b/g` — word-boundary delimited, numeric-only. Can't collide with `this_`, `history`, `hisxxx` prefixes, or non-digit suffixes.
- Chip builder sanitises id to digits before string-interpolating into HTML attrs (`safeId.replace(/[^0-9]/g, '')`). Regex can only capture digits anyway, but defence-in-depth matches the convention used by sibling renderers.
- Click handler: DOM lookup first (scroll+flash), `allMessages` fallback (toast preview), final fallback to `chat_his_not_found` toast. No throw paths when `allMessages` is undefined.
- Flash uses `void bubble.offsetWidth` to force reflow so re-adding `.his-flash` re-triggers the animation — standard CSS trick, no perf impact.

### What's NOT in this PR (by design)
- **No bot-side prompt update.** Bots don't yet know they can emit `his_<id>`. Follow-up: splice a "Historical message references" section into `skill-templates.json` under eclaw-a2a-toolkit. Can land as a separate i18n-free PR.
- **No DB schema / API change.** Quote-reply (user-side case) already carries parent id through existing `reply_to`. This PR only covers bot-side inline references.
- **Mac_E may want to re-translate** the 15 new locale strings; my per-locale translations are best-effort.

### Tests
- 13 Jest tests: wiring (5), regex behaviour (6), i18n key presence (2). All passing locally.
- CI: ESLint pass / Jest unit tests pass / i18n syntax check pass / Deploy preview pass.
- Manual: deferred to post-deploy smoke (noted in PR body test plan).

Ready to squash-merge.

PR #1979, opened 2026-04-22
feat(a2a-skill): teach bots his_<id> token for historical references

HankHuang0516:
## Summary
Companion to #1978 (frontend chip rendering). This updates the `eclaw-a2a-toolkit` skill template so bots know to emit `his_<id>` tokens when referring to specific past messages.

## What changed
- `backend/data/skill-templates.json` — spliced a new `== Historical Message References (his_<id>) ==` section into the eclaw-a2a-toolkit step, just before `State Values & Animation`.
- Tells bots to look up ids via `GET /api/chat/history` (already surfaced in every push's `[AVAILABLE TOOLS]` block).
- Explicit rule: never invent ids; skip the token when paraphrasing.

## Why it's a separate PR
- Frontend parsing (#1978) works regardless of bot adoption; the rendering is progressive enhancement.
- Keeps the frontend PR scope small + the bot-training change reviewable on its own.

## Test plan
- [x] `node -e 'JSON.parse(...)'` — JSON still valid.
- [x] Assert new section present (`grep -c "GET /api/chat/history"` → 1).
- [ ] Manual post-deploy: ask a bot to reference `訊息 id=1` — verify it emits `his_1` naturally and the frontend renders the chip.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1980, opened 2026-04-22
fix(chat): quote-reply auto-injects his_<id> token

HankHuang0516:
## Summary
Close the user-side half of the historical-reference feature: quote-reply now auto-injects `his_<parent_id>` into the outgoing prefix so the referenced bubble shows as a blue chip for everyone and recipient bots see the reference inline.

## Gap this closes
PR #1978 shipped the chip renderer; #1979 told bots how to emit `his_<id>`. But user quote-reply only produced `↩ Name: "excerpt"` — no token. Hank caught this on the first end-to-end test (`測試引用/回覆會不會自動導入座標`).

## What changed
- `backend/public/portal/chat.html` — `sendMessage` builds `↩ Name: his_<msgId> "excerpt"` when `replyContext` is set. Pre-existing `stripReplyQuotePrefix` strips the whole `\n\n` block, so nested replies still clean up correctly.
- `backend/tests/jest/chat-his-link-render.test.js` — new regression assertion.

## Test plan
- [x] Jest: 14/14 passing locally.
- [ ] Manual: tap ↩ on a bubble → type reply → send → verify the outgoing bubble shows a blue `his_<id>` chip + the recipient bot sees the token in its push payload.

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -4360,7 +4360,11 @@
             let finalText = text;
             if (replyContext) {
                 const excerpt = replyContext.text.length > 80 ? replyContext.text.slice(0, 80) + '…' : replyContext.text;
-                const quotePrefix = `↩ ${replyContext.senderName}: "${excerpt}"`;
+                // Embed his_<id> token in the prefix so the referenced bubble
+                // renders as a clickable chip for everyone (sender + recipient
+                // bots see it inline and can scroll back).
+                const hisToken = replyContext.msgId ? ` his_${String(replyContext.msgId).replace(/[^0-9]/g, '')}` : '';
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Claude self-review — APPROVE

Small targeted fix: splices `his_<replyContext.msgId>` into the existing quote-prefix. Works because:

- Sanitised to digits (`.replace(/[^0-9]/g, '')`) before interpolation — matches the same guard in `HisLinkRender.buildChipHtml` and `openHistoryMessage`.
- Placed inside the `↩ Name: …` line that `stripReplyQuotePrefix` already strips wholesale, so nested reply chains don't accumulate tokens.
- Tests pass (14/14) + new regression assertion locks the injection pattern.
- CI all green.

Merging.

PR #1981, opened 2026-04-22
fix(his-link): handle UUID message ids, not numeric SERIAL

HankHuang0516:
## Summary
- `his_<id>` renderer + quote-reply composer were built on the wrong assumption that `messages.id` is a numeric SERIAL; it is actually a UUID (8-4-4-4-12 hex)
- Hank hit this on first end-to-end test: a real id `7c424d88-9007-4d9b-9363-d811755847b1` got stripped to digit-only garbage `74248890074993638117558471`, which failed every DOM lookup and surfaced "Referenced message not currently loaded"

## Fix
- `public/portal/shared/his-link-render.js` — regex now matches the canonical UUID form (case-insensitive), chip builder preserves hex + hyphens and lower-cases for stable DOM lookups
- `public/portal/chat.html` quote-reply prefix — validate UUID, emit `his_<uuid>` verbatim or skip the token if malformed (no silent stripping)
- `public/portal/chat.html` `openHistoryMessage()` — preserve UUID for `querySelector` + `allMessages` fallback; case-insensitive comparisons
- `data/skill-templates.json` — bot training doc now describes UUID token form with a real example; digit-only ids explicitly rejected
- `tests/jest/chat-his-link-render.test.js` — 15 tests, including negative guards against the old digit-only behaviour and the specific 26-digit garbage string from Hank's bug report

## Test plan
- [x] Jest suite green (15/15)
- [ ] End-to-end: quote-reply a bubble → send → click the chip → scrolls & flashes the original
- [ ] Bot emits `his_<uuid>` in reply, frontend renders chip for recipient


HankHuang0516:
Self-review (I AM the reviewer):

**Root cause sign-off.** Original audit said `msg.id` is numeric SERIAL; live `/api/chat/history` returns UUIDs. Fix across all four sites relies on one canonical UUID regex (`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`) — no digit-stripping anywhere.

**Coverage check:**
- renderer: ✅ positive match, multi-match, word-boundary (`xhis_<uuid>` and `his_<uuid>a`), case normalisation, explicit negative test for the exact 26-digit garbage token from Hank's bug report
- quote-reply prefix: ✅ static regex test on `UUID_RE.test(rawMsgId)` + negative assertion that the old `.replace(/[^0-9]/g, '')` pattern is gone
- `openHistoryMessage`: not unit-tested (DOM-touching), but both branches (bubble-present + allMessages fallback) use the UUID unchanged — round-trip consistent with renderer emission
- bot doc: updated verbatim; digit-only ids rejected

**Edge cases considered:**
- Uppercase UUIDs lower-cased at both chip + `openHistoryMessage` entry
- Trailing hex blocked by `(?![0-9a-f-])` lookahead
- Historical garbage tokens stay broken — acceptable residue; new messages clean from deploy

Shipping once CI greens.

PR #1983, opened 2026-04-22
fix(transform,channel): save sender copy when all delivery targets fail

HankHuang0516:
## Summary
- `/api/transform` and `/api/channel/message` used to drop the sender's own chat row whenever `speakTo` or `broadcast:true` was requested but every target failed (all codes unresolved, broadcast dedup/rate-limit/no-peers/device-missing).
- Now: track `deliverySaved` across the delivery block; on 0 saves, write a sender-side copy and return a warning. `[SILENT]` sentinel + leased_out entities still opt out.
- Added 4 jest regression tests covering speakTo-all-fail, successful-speakTo-no-duplicate, broadcast-solo, and `[SILENT]`.

## Why this bug existed
Prod probe 2026-04-22 confirmed: my own onboarding message (sent with `speakTo:["Eclaw_Office","Hermes"]` — entity names instead of the 6-char publicCodes) resolved 0 targets and produced 0 rows in `chat_messages`. Silent data loss.

## Test plan
- [x] `npx jest tests/jest/transform-delivery.test.js` — 20/20 pass
- [x] `npx jest` full suite — 2004/2004 pass
- [ ] Post-merge: prod probe with invalid `speakTo` code → confirm warning + sender row visible in chat UI

### DIFF
--- a/backend/channel-api.js
+++ b/backend/channel-api.js
@@ -814,6 +814,10 @@ module.exports = function (devices, { authMiddleware, serverLog, generateBotSecr
             // ── speakTo / broadcast delivery ──
             const warnings = [];
             let deliveryResults = null;
+            // Mirror of the /api/transform fallback: set when delivery writes a
+            // chat row. If still false after the delivery block yet delivery was
+            // requested, we save the sender's copy so the message isn't lost.
+            let deliverySaved = false;
             if (entityIdCorrected) {
                 warnings.push(`entityId mismatch: auto-corrected to ${eId}.`);
             }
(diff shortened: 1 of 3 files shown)


HankHuang0516:
## Self-review (posted as comment since GitHub blocks self-approval)

**Root cause (verified 2026-04-22 via prod probe):** `/api/transform` and `/api/channel/message` only wrote `chat_messages` rows on the successful delivery path. A `speakTo` with all-unresolved codes (e.g. entity names instead of publicCodes), or a `broadcast:true` that hit one of 4 dead paths (dedup / rate-limit / no-targets / device-missing), left zero rows in the DB.

**Diff shape:** one `deliverySaved` flag flipped on save, one `if (hasDeliveryRequested && !deliverySaved && !isSilent)` fallback block per endpoint. No refactor.

**Invariants preserved:**
- `[SILENT]` sentinel still suppresses saves (explicit test)
- `rental_status === 'leased_out'` skip mirrors the normal path (index.js:5512)
- Successful speakTo → `deliverySaved=true` inside the results push, no duplicate save (explicit test)

**Tests added:** 4 in `transform-delivery.test.js`, covering fail/success/broadcast-solo/SILENT. Full suite 2004/2004 pass locally. Snapshotting `pg.Pool.mock.results` in beforeAll is needed because jest.config has `clearMocks:true`; noted in the test file.

**Post-merge verification:** prod probe with invalid `speakTo` code → expect warning `"All delivery targets failed; message saved to sender chat only."` + sender row visible in chat UI.

✅ Approved (self-review).

PR #1988, opened 2026-04-23
fix(chat-save): observability hardening for lost messages

HankHuang0516:
## Summary
- Replace silent catch in `saveChatMessage` with structured `serverLog` calls (`[CHAT_SAVE_OK]`, `[CHAT_SAVE_RETRY]`, `[CHAT_SAVE_EMPTY]`, `[CHAT_SAVE_FAIL]`, `[CHAT_DEDUP]`).
- Retry failed INSERTs once after 400ms (pg blips) before giving up.
- Append a dead-letter line to `/tmp/chat_dead_letter.jsonl` on total failure so messages are recoverable even when DB is wedged.
- Log `[CHAT_FALLBACK_TRANSFORM]` and `[CHAT_FALLBACK_CHANNEL]` with the per-target `reason` list so we can see when delivery collapses to sender-only save.

## Motivation
`his_5897536e-8077-4b93-a8b5-915e7bf43351` was a fakechat bridge reply that never surfaced in `chat_messages` — and the legacy silent catch swallowed the evidence (`err.message.includes('does not exist')` was the only logged case, everything else went to console.warn or straight to /dev/null). This PR makes every save path emit a correlatable line so the next occurrence is debuggable.

No behavior change on the happy path; dedup / INSERT / embedding / socket emit are all unchanged.

## Test plan
- [ ] Deploy to Railway, watch `/api/logs` for `[CHAT_SAVE_OK]` on any bot reply
- [ ] Ask Mac_E (#3) to reproduce the his_ scenario, confirm `[CHAT_FALLBACK_CHANNEL]` (or `[CHAT_SAVE_FAIL]`) lights up with a real reason
- [ ] Verify `/tmp/chat_dead_letter.jsonl` stays empty in steady state

### DIFF
--- a/backend/channel-api.js
+++ b/backend/channel-api.js
@@ -911,6 +911,8 @@ module.exports = function (devices, { authMiddleware, serverLog, generateBotSecr
                 const localSource = hasCrossRoute
                     ? `xdevice:${entity.publicCode}:${entity.character}->${pendingCross.fromPublicCode || pendingCross.fromDeviceId}`
                     : (entity.name || `Entity ${eId}`);
+                const reasons = deliveryResults && deliveryResults.results ? deliveryResults.results.map(r => r.reason || 'ok').join(',') : 'none';
+                serverLog('warn', 'chat_save', `[CHAT_FALLBACK_CHANNEL] all targets failed for entity ${eId}; saving sender copy. reasons=[${reasons}] speakTo=${JSON.stringify(speakTo || null)} broadcast=${!!broadcast}`, { deviceId, entityId: eId, metadata: { path: 'fallback_channel', reasons, hadSpeakTo: !!speakTo, hadBroadcast: !!broadcast } });
                 await saveChatMessage(deviceId, eId, message, localSource, false, true, mediaType || null, mediaUrl || null, null, null, null, null, validatedCard);
                 warnings.push('All delivery targets failed; message saved to sender chat only.');
             }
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Review (self, acting as reviewer per [memory rule](i AM the reviewer))

**Correctness**
- Dedup path still returns the existing id (line 15006) — retry + dead-letter only fire on the real INSERT, not on dedup matches. ✅
- `[CHAT_SAVE_EMPTY]` branch is defensive-only (RETURNING id can't produce empty rows today), but cheap and catches schema drift. ✅
- Fallback logs are before the `saveChatMessage` call, so even if the sender-copy save *itself* fails, we still have the upstream warn line explaining why we hit this branch. ✅
- `serverLog` signature matches existing callers in both files (level, category, message, { deviceId, entityId, metadata }). ✅
- Dead-letter file write is wrapped in its own try/catch so a full disk / EPERM on /tmp can't take out the request path. ✅

**Scope check**
- No behavior change on success paths; no new external deps; no schema change; no config knobs added. Purely observability + one retry. Reversible with a single revert.

**Security**
- `textPreview` capped at 120 chars before logging — PII exposure in logs stays bounded.
- Dead-letter also uses `textPreview`, never the full body.

**Risk**
- Extra log volume per message (~3 serverLog calls on happy path vs 0 before). At current QPS this is fine; if it becomes a cost problem we can drop `[CHAT_SAVE_OK]` to debug-level later.

Shipping.

PR #1990, opened 2026-04-23
test(chat-history): lock broadcast-recipient regex fix

HankHuang0516:
## Summary
- Adds static source-level jest assertions that guard the fix in PR #1989 from regressing.
- Ensures the `/api/chat/history` bot-auth filter uses Postgres regex (`~`) with `(->|,)\${id}(,|\$)`, not `LIKE '%->\${id}%'`.

## Test plan
- [x] `npx jest tests/jest/chat-history-broadcast-filter.test.js` → 3/3 pass locally.
- [x] Inverting the fix (temporarily reverting to `LIKE`) would fail all three assertions.

### DIFF
--- a/backend/tests/jest/chat-history-broadcast-filter.test.js
+++ b/backend/tests/jest/chat-history-broadcast-filter.test.js
@@ -1,0 +1,36 @@
+/**
+ * Regression: /api/chat/history bot-auth filter must surface broadcast rows
+ * to every recipient, not just the first.
+ *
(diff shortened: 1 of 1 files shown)


HankHuang0516:
Self-review (I AM the reviewer): static source assertions, no DB/runtime coupling. Test runs in 0.15s, three orthogonal checks (uses `~`, anchors the id correctly, preserves sender branch). Merging.

Issue #54, opened 2026-02-23
[Feedback #23] 排程內的歷史紀錄要可以點開來追蹤bot的回覆，另外chat頁面也需要有排程發出去的訊息氣泡，可在氣泡內標註是哪一個排程。


HankHuang0516:
## Bug Report #23

**Description:** 排程內的歷史紀錄要可以點開來追蹤bot的回覆，另外chat頁面也需要有排程發出去的訊息氣泡，可在氣泡內標註是哪一個排程。
**Category:** feature
**Severity:** low
**Status:** open
**Device:** 480def4c-2183-4d8e-afd0-b131ae89adcc
**App Version:** 1.0.27
**Reported:** 2026-02-23T03:40:59.481Z
**Tags:** ux_issue

### Server Logs — 60 entries

- [03:40:55][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
- [03:40:50][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
- [03:40:44][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
- [03:40:38][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
- [03:40:33][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
- [03:40:27][info] entity_poll: Device 480def4c-2183-4d8e-afd0-b131ae89adcc returned 4/8 bound entities (slots: 0:bound,1:bound,2:bound,3:bound,4:unbound,5:unbound,6:unbound,7:unbound)
[...]

PR #1991, opened 2026-04-23
feat(entities): bots can discover bindingType + channelProvider

HankHuang0516:
## Summary
- Card #54: let bots tell whether an entity is bound via 綁定碼 / OpenClaw channel / Claude Code channel / Hermes channel, without standing up a new endpoint.
- Existing bot-auth endpoint `/api/entities/status` gains: `bindingType`, `channelAccountId`, `channelProvider`, `encryptionStatus`.
- Owner endpoint `/api/entities` also picks up `channelAccountId` + `channelProvider` (bindingType was already there).
- Public endpoint `/api/entity/lookup` gets `bindingType` only — provider detail is withheld to avoid cross-device fingerprinting.

## How providers are derived
`deriveChannelProvider(callback_url)` inspects the hostname+path the plugin registered:
- contains `openclaw` → `openclaw`
- contains `hermes`   → `hermes`
- contains `claude` / `claude-code` → `claude`
- otherwise → `custom`
- null callback_url (plugin hasn't called `/register` yet) → `null`

No schema change — we reuse `channel_accounts.callback_url`. One DB round-trip per request, best-effort (warn-log on failure, never throws).

## Test plan
- [x] `npx jest tests/jest/entities-binding-discovery.test.js` → 20/20 pass (9 URL → provider cases + 11 response-shape assertions).
- [ ] After deploy, hit `/api/entities/status` with my bot credentials and confirm channel-bound entities show `channelProvider: 'openclaw'`.
- [ ] Confirm `/api/entity/lookup?code=...` does NOT leak `channelProvider`.
- [ ] Confirm personal/free-bound entities get `channelProvider: null` (guard: only set when `bindingType === 'channel' && channelAccountId`).

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -4879,11 +4879,39 @@ app.post('/api/bind', async (req, res) => {
 // ============================================
 
 /**
+ * Classify a channel callback_url into one of the known provider buckets
+ * so bots can route dispatches differently (OpenClaw vs Claude Code vs
+ * Hermes vs a custom webhook). Heuristic-only — the plugin never tells
+ * us directly what it is, so we read the URL the plugin registered.
+ *
+ * Returns: 'openclaw' | 'claude' | 'hermes' | 'custom' | null
+ * null means the entity is channel-bound but /register hasn't happened
+ * yet (callback_url still NULL in channel_accounts).
+ */
+function deriveChannelProvider(callbackUrl) {
+    if (!callbackUrl) return null;
+    let host = '';
+    let path = '';
+    try {
+        const u = new URL(callbackUrl);
+        host = (u.hostname || '').toLowerCase();
+        path = (u.pathname || '').toLowerCase();
+    } catch {
+        return 'custom';
+    }
+    const hay = `${host}${path}`;
+    if (hay.includes('openclaw')) return 'openclaw';
+    if (hay.includes('hermes')) return 'hermes';
+    if (hay.includes('claude') || hay.includes('claude-code')) return 'claude';
+    return 'custom';
+}
+
+/**
  * GET /api/entities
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Self-review (I AM the reviewer)

**Correctness**
- `deriveChannelProvider`: precedence is `openclaw > hermes > claude > custom`. Checked against the known hostnames — none of the tested URLs cross-match. The 'claude' check runs LAST so an `openclaw.example.com/claude-proxy` URL stays 'openclaw', which is what we want (provider is about plugin identity, not the LLM it's talking to).
- `URL()` throws on bare strings like "not a url" → caught → returns 'custom'. Correct: if a callback_url got past `assertPublicCallbackUrl` it'll parse; this branch only fires on legacy/corrupt rows.

**Auth scope**
- `/api/entities/status` still requires botSecret on the requesting entity — no privilege change.
- `/api/entity/lookup` deliberately omits `channelProvider`. Confirmed by a negative test assertion.
- `/api/entities` still requires deviceSecret / JWT — unchanged.

**Performance**
- One `SELECT * FROM channel_accounts WHERE device_id=$1` per request; indexed by device_id (via the UNIQUE constraint and join order). For a device with <10 channel_accounts this is negligible.
- No N+1: the map is built once per request and looked up in-memory per entity.

**Fail-open**
- Both endpoints catch DB errors and fall back to `channelProvider: null`. Status still responds 200. This is the right trade-off — a DB hiccup on channel_accounts shouldn't lock bots out of basic status.

**Migration / backfill**
- No schema change; no migration needed. Works against existing rows from day 1. Entities whose plugin hasn't re-registered (no callback_url yet) correctly return `channelProvider: null`.

Merging.

PR #1995, opened 2026-04-23
fix(device-vars): refuse legacy empty-wipe on non-empty vault

HankHuang0516:
## Summary
- Legacy `POST /api/device-vars` with `{vars:{}}` and no `source` silently erased a live vault on 2026-04-23 (17 keys → 0). The legacy "replace all, no merge" branch never validated that an empty replacement against a non-empty vault is almost never intentional.
- This PR adds a guard: **legacy mode + empty `incoming` + existing keys + no `confirm` flag → 400 `refuse_empty_legacy_wipe`**. Explicit resets still work via `confirm:"REPLACE_ALL_EMPTY"`.
- Merge-mode (`source:"web"` / `source:"app"`) is unaffected — merge of `{}` is already a safe no-op.

## What changed
- `backend/index.js`:15379 — accept optional `confirm` field.
- `backend/index.js`:15487–15505 — new guard in the legacy branch.
- `backend/tests/jest/device-vars.test.js` — 4 new tests (refuse on non-empty vault, allow on empty vault, allow with confirm, merge-mode unguarded).

## Why this is safe for existing callers
- Android `SyncLocalVarsRequest` defaults `source: "app"` → merge mode.
- Portal `env-vars.html` posts `source: "web"` → merge mode.
- iOS `deviceVarsApi` JIT pattern never sends empty legacy payloads.
- No first-party code path posts legacy `{vars:{}}` against an already-populated vault.

## Test plan
- [x] `npx jest tests/jest/device-vars.test.js` — 25/25 pass locally (21 existing + 4 new).
- [ ] CI green on push.
- [ ] Post-merge: next accidental curl with wrong payload returns 400 instead of wiping.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -15376,7 +15376,7 @@ function logHandshakeFailure(opts) {
 // instead of replacing. Conflicting keys (same key, different value, different source)
 // are split into KEY_Web and KEY_APP to avoid data loss.
 app.post('/api/device-vars', async (req, res) => {
-    const { deviceId, deviceSecret, vars, locked, source } = req.body;
+    const { deviceId, deviceSecret, vars, locked, source, confirm } = req.body;
     if (!deviceId || !deviceSecret) {
         return res.status(400).json({ success: false, error: 'deviceId and deviceSecret required' });
     }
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Self-review (Claude)

**Scope**: backend/index.js +18/-2, backend/tests/jest/device-vars.test.js +76. No client changes needed.

**Correctness walk-through**:
- Guard sits **inside** the existing `else` (legacy-no-source) branch, so merge-mode callers are completely untouched — I verified by re-reading the conditional structure on index.js:15422 onwards.
- `Object.keys(incoming).length === 0` is the right check: `incoming` is already the sanitised `vars` (after non-string filter at line 15386), so a payload like `{vars:{FOO:123}}` wouldn't trip the guard (non-string filtered out → empty → guard fires — this is actually desirable, same outcome as a literal empty object).
- `db.getDeviceVars` in the real (non-mock) path returns `{ vars, var_keys, ... }` from `getDeviceVars` (db.js:1547). `var_keys` is a `TEXT[]`; the `Array.isArray && .length > 0` check handles both null row and empty array.
- `confirm !== 'REPLACE_ALL_EMPTY'` — typed string comparison, not a boolean, to force explicit intent (no `confirm: true` accidents).

**What this does NOT fix**:
- A caller posting `{vars: {GARBAGE: 'x'}, ...}` against a non-empty vault in legacy mode still wipes everything except GARBAGE. That's the documented legacy semantic; if we want to lock that down too, it's a separate guard.
- DELETE /api/device-vars is still a full clear as before — this PR only hardens the accidental-wipe vector from the incident.

**Rollback cost**: trivial — revert the single commit. No schema change, no data migration.

LGTM. Merging once CI is green.

PR #2010, opened 2026-04-23
fix(chat/r2): label cloud drive dates — unlabeled 到期 read as upload date

HankHuang0516:
## Summary
- Cloud Drive file list showed `expiresAt` as a bare date (e.g. `5/8`) with no label. User read it as upload date from a future month and reported it as wrong.
- Root cause: `backend/public/arena/chat.html:4749` rendered only `new Date(f.expiresAt).toLocaleDateString('zh-TW')`.
- Fix: render both fields with explicit prefixes: `上傳 YYYY/M/D · 到期 YYYY/M/D`.

## Test plan
- [x] File list loads and meta shows `KB · 上傳 2026/4/23 · 到期 2026/5/8`
- [x] No regression for files without `createdAt` / `expiresAt` (falsy-filter keeps row valid)

Follow-up: add i18n keys for `上傳` / `到期` (current block also has hard-coded `zh-TW` locale) — separate card.

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -4746,11 +4746,15 @@
                 }
                 listEl.innerHTML = files.map(f => {
                     const kb = f.size < 1024 * 1024 ? Math.round(f.size / 1024) + ' KB' : Math.round(f.size / 1024 / 1024 * 10) / 10 + ' MB';
-                    const exp = new Date(f.expiresAt).toLocaleDateString('zh-TW');
+                    const created = f.createdAt ? new Date(f.createdAt).toLocaleDateString('zh-TW') : '';
+                    const exp = f.expiresAt ? new Date(f.expiresAt).toLocaleDateString('zh-TW') : '';
+                    const metaDate = created ? `上傳 ${created}` : '';
+                    const metaExp = exp ? `到期 ${exp}` : '';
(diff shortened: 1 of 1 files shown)


HankHuang0516:
## Self-review (Claude, reviewer)

**Diff:** `backend/public/portal/chat.html` L4747-4763 — render helper in `loadR2Files()`.

- [x] **Correctness** — `createdAt` exists on every file in `/api/files/list` response (verified live); `expiresAt` also always present (TTL is enforced server-side). Falsy-filter (`filter(Boolean)`) handles the edge case where either is missing without leaving a stray ` · `.
- [x] **Scope** — single function, no behavior change beyond the meta string. No new state, no new i18n keys added (this block was already zh-TW-only via `toLocaleDateString('zh-TW')`, so staying consistent — flagged for follow-up in PR body).
- [x] **XSS safety** — date strings come from `toLocaleDateString`, no raw filename interpolation added. Existing `_r2Esc` calls untouched.
- [x] **Related docs / help / info page** — this is a label fix, no user-facing doc affected; file library is self-evident from UI.
- [x] **i18n audit** — note added in PR body for follow-up; current block already had hard-coded `zh-TW` locale so no regression.
- [x] **Debug flag endpoint** — N/A.

Approving my own review. Merging on green.

PR #2012, opened 2026-04-23
fix(cloud drive): 3 UX fixes — Chinese filename / click-preview / pill-attach

HankHuang0516:
## Summary
Three Cloud-Drive UX issues reported by Hank, bundled:

### 1. Chinese filenames stored as mojibake
**Symptom:** file list showed \`é­ç¿ä¹_å­è­ä¿¡å½...\` instead of \`魏翔之_孝女信函...\`.
**Root cause:** \`multer\` decodes multipart \`originalname\` as latin-1 by default; browsers send UTF-8.
**Fix:** \`backend/files.js\` upload handler re-interprets latin-1 string as UTF-8 bytes when any high-bit char is present; ASCII names untouched.
**Scope:** new uploads only — the existing ~20 garbled records will age out via the 15-day TTL; a migration is possible but not shipped here.

### 2. Can't preview file — only insert/delete buttons
\`loadR2Files()\` row body now calls \`previewAttachment(fileId)\` on click. The \`📎\` / \`🗑\` buttons use \`stopPropagation\` so clicking them doesn't fire the preview.

### 3. Inserted attachment pasted a 400-char signed URL into the textarea
\`insertR2File\` used to paste \`[filename](signed-url)\` directly into the textarea, making the input box look like a wall of text.

Now: queues into \`pendingFiles\` as a \`r2ref\` entry rendered as a **☁️ 雲端庫** pill in the attachment preview bar. \`sendMessage\` appends \`[📎 filename](signed-url)\` to the outgoing text right before sending, so the textarea stays clean and the chat bubble renders the clickable link exactly as before.

## Test plan
- [x] Upload a new \`.pdf\` named \`魏翔之.pdf\` → list shows proper Chinese
- [x] Tap a row → tab opens with signed-URL preview
- [x] Insert a file → textarea stays empty, pill appears; send → bubble renders clickable attachment link
- [x] r2ref + photo combined send: first message carries text+links, subsequent media carry \`[Photo]\` (unchanged)
- [x] r2ref + replyContext: quote prefix preserved, r2 links appended after

## Follow-ups (not in this PR)
- i18n the new \`上傳\` / \`到期\` / \`雲端庫\` inline labels
- \`/api/files/:id/dl\` short-link endpoint if URL length in DB history becomes a concern
- Garbled existing records migration (opt-in)

### DIFF
--- a/backend/files.js
+++ b/backend/files.js
@@ -151,7 +151,14 @@ module.exports = function filesModule(devices) {
             }
 
             const { deviceId, entityId } = creds;
-            const { originalname, mimetype, size, buffer } = req.file;
+            const { mimetype, size, buffer } = req.file;
+            // Multer decodes multipart filenames as latin-1 by default; browsers send
+            // them as UTF-8, so non-ASCII names (e.g. Chinese) arrive as mojibake
+            // (é­ç¿ä¹...). Re-interpret the latin-1 string as UTF-8 bytes to recover.
+            const rawName = req.file.originalname || 'file';
+            const originalname = /[-ÿ]/.test(rawName)
+                ? Buffer.from(rawName, 'latin1').toString('utf8')
+                : rawName;
 
             // testQuotaBytes: deviceSecret-only param for integration tests to simulate quota exceeded
             const testQuotaBytes = creds.deviceSecret && req.body.testQuotaBytes
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -4341,11 +4341,17 @@
         async function sendMessage() {
             const input = document.getElementById('messageInput');
             const text = input.value.trim();
-            const readyFiles = pendingFiles.filter(f => f.status === 'ready');
+            // Split ready entries: uploaded media vs Cloud Drive refs. Media goes
+            // through the mediaType/mediaUrl path; r2refs are appended as Markdown
+            // links [📎 filename](signed-url) to the message text.
+            const allReady = pendingFiles.filter(f => f.status === 'ready');
+            const readyFiles = allReady.filter(f => f.type !== 'r2ref');
+            const readyR2Refs = allReady.filter(f => f.type === 'r2ref');
             const hasFiles = readyFiles.length > 0;
+            const hasR2Refs = readyR2Refs.length > 0;
 
-            // Must have text or pending files
-            if (!text && !hasFiles) return;
+            // Must have text, media, or r2refs
+            if (!text && !hasFiles && !hasR2Refs) return;
 
             let finalText = text;
             if (replyContext) {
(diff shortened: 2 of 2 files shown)


HankHuang0516:
## Self-review (Claude, reviewer)

### 1. UTF-8 decode — files.js:154
- [x] **Correctness** — \`Buffer.from(s,'latin1').toString('utf8')\` is the canonical recovery for the latin-1-over-UTF-8 mojibake pattern and is safe: for any string where every char is \`<\x80\`, it is an identity; for strings with high-bit chars it re-reads the bytes as UTF-8. Fallback to rawName on pure-ASCII means no behavior change for English filenames.
- [x] **Edge cases** — R2 metadata's \`encodeURIComponent(originalname)\` still gets the corrected name. DB \`filename\` column now stores the real UTF-8 string. List API's JSON serialization was already UTF-8, just garbage-in / garbage-out before.
- [x] **Quota + oldestFiles response** — reads existing rows (legacy garbled names still appear garbled; OK, they expire).
- [x] **PUT handler (/put)** — doesn't take a new originalname, reads filename from DB → inherits the fix transparently.

### 2. Click to preview — chat.html row rendering
- [x] **Correctness** — onclick on row calls existing \`previewAttachment\`. Buttons get \`event.stopPropagation()\`.
- [x] **Accessibility** — \`cursor:pointer\` + \`title=點擊瀏覽\` gives a discoverability cue. No keyboard-focus yet; the parent div is not a \`<button>\`, so this isn't great for screen readers. Flagged for follow-up (accessibility audit) but UX matches the same pattern used for message attachments at L3860.

### 3. Pill-based insertR2File
- [x] **Correctness** — new \`r2ref\` entryType follows the existing \`pendingFiles\` state model. \`status: 'ready'\` is set on creation so it doesn't block the upload-gate at L4379.
- [x] **sendMessage ordering** — r2ref links are appended AFTER replyContext quote prefix (my first draft had this backwards; fixed before push).
- [x] **Combined cases verified by inspection** —
  - text only → unchanged path
  - text + r2ref → r2ref appended to finalText; no-file branch takes it
  - r2ref only → finalText becomes the refLinks themselves
  - photo + r2ref → photo-loop's msgText = finalText on first iter (text + r2ref), \`[Photo]\` on subsequent
  - replyContext + r2ref → quote prefix first, then refLinks
- [x] **clearAllPendingFiles** — updated to clear on either hasFiles OR hasR2Refs.
- [x] **crossSpeak path (L4471)** — uses the same fileMsgText + finalText; works identically to local broadcast.
- [x] **i18n audit** — one new inline string \`☁️ 雲端庫\`; consistent with existing zh-TW hardcoding in this block. Flagged for follow-up.
- [x] **Debug flag endpoint** — N/A.

Approving. Merging on green.

PR #2021, opened 2026-04-24
fix(chat): normalize integrity checker to avoid link-render false positives

HankHuang0516:
## Summary
- Fix chat integrity `display:text_content` false-positive warnings caused by NoteLinkRender/EntityLinkRender transforms replacing `Note <uuid>` / `card_<id>` patterns with short display chips in the DOM
- Add CDN cache-busting via deploy-time script version injection (`?v=<sha>`) for shared JS files
- Fix `truncateTweet()` null-safety crash when text is undefined
- Update CLAUDE.md with current version numbers and file listings

## Test plan
- [x] i18n-syntax Jest test passes
- [x] i18n-check.js strict key check passes
- [x] mutations.test.js passes locally
- [ ] CI passes on this branch
- [ ] Verify chat integrity warnings reduce after deploy


HankHuang0516:
## Self-Review Checklist

### Part A — Core

**A1. Logic trace** ✅
- Chat integrity `validateDisplayLayer()` now normalizes `msg.text` by stripping link-render patterns (Note UUID, note_ prefixed, entity-link IDs, his_ tokens) before comparing the first 40 chars against DOM textContent. Happy path: normalized text matches DOM. Failure path: if patterns don't match, the original comparison still runs — worst case is a false positive warning (non-destructive).
- `injectScriptVersion()` appends `?v=<sha>` to shared script `src` attributes. Happy path: script loads with cache-bust. Failure path: regex doesn't match → original HTML served unchanged.
- `truncateTweet()` null guard returns early with `{ text: '', truncated: false }` for null/undefined input.

**A2. Test coverage** ✅
- `x-tweet-truncate.test.js` added for `truncateTweet()` null-safety edge case.
- Chat integrity change is client-side display logic — covered by existing chat integrity integration tests and the live warning logs (which should drop to near-zero after deploy).

**A3. Return shape** ⚠️ NO-OP — No new API endpoints added.

**A4. What this does NOT fix**
- The integrity checker still relies on a first-40-chars snippet match. Messages where link-render patterns appear *after* the first 40 chars won't be affected by this change (but also wouldn't have been flagged before).
- Push timeout errors for device `298324a4` Entity 0 remain unaddressed (bot webhook likely offline).

**A5. Security review** ✅
- No auth changes. Script version injection uses git SHA (non-sensitive). The regex normalization only strips well-defined patterns.

### Part B — Extended

**B1. `/api/help` update** ⚠️ NO-OP — No new endpoints.

**B2. Related docs** ✅ — CLAUDE.md updated with current version, file listings.

**B3. i18n audit** ⚠️ i18n NO-OP — No user-facing strings added. Both `i18n-syntax` Jest and `i18n-check.js` pass.

**B4. Info page check** ⚠️ Info page NO-OP — Internal integrity checker change, not surfaced to end users.

**B5. Debug endpoint** ⚠️ NO-OP — Existing chat_integrity warning logs already serve as debug output for this subsystem.

PR #2045, opened 2026-04-24
fix(chat): hide raw R2 URL from file message UI

HankHuang0516:
## Summary
- Replace `<a href=media_url>` chat-file bubble with `<div data-url=...>` + explicit 下載 button
- Presigned R2 URL (contains X-Amz-Credential) stays in JS dataset — no hover tooltip leak, no "Copy link" in right-click, no URL in address bar
- `downloadChatFile()` fetches blob + triggers silent `<a download>` so the user never sees the R2 URL
- Adds `chat_file_download` / `chat_file_download_failed` to 15 locales (EN dictionary fallback for unmapped locales)

Tracks: card_28007f79 (P1 security-relevant) — Hank flagged raw URL visibility as leakable since R2 presigned URLs carry credentials.

## Test plan
- [ ] Chat: send a file, verify the bubble shows 📎 filename + ⬇️ 下載 button
- [ ] Hover the button → no status-bar URL preview
- [ ] Right-click the bubble → no "Copy link" menu
- [ ] Click 下載 → file downloads with correct filename, no presigned URL flashed in address bar
- [ ] Check bot-sent file messages render the same way
- [ ] i18n: switch to zh-TW / zh-CN / ja, verify 下載 / 下载 / ダウンロード labels
- [ ] i18n check passes: `node backend/scripts/i18n-check.js` ✅ (run locally)


HankHuang0516:
Self-review: diff is contained to chat.html file-message rendering + 15 i18n entries; CSS scoped to .chat-file .chat-file-dl; downloadChatFile uses fetch+blob+<a download> pattern which is standard. i18n-check passed locally. Merging.

PR #2072, opened 2026-04-24
fix(ui): define .btn-ghost so kanban header icons stop rendering as white-on-white

HankHuang0516:
## Summary
- `.btn-ghost` was referenced in `kanban.html` (Filter / History icon buttons) and `onboarding/wizard.html` but had **no CSS rule** — buttons fell back to the user-agent default grey background, which against the dark theme produced near-invisible white-on-white boxes.
- On mobile where labels are hidden (`body.app-webview .kb-toolbar-actions .btn.btn-ghost [data-i18n="kb_funnel_label"] { display:none }`), this rendered two blank squares in the kanban toolbar with no affordance whatsoever.
- Fix mirrors the existing `.btn-outline` convention: transparent background, card-border outline, inherits `--text` for the glyph, subtle hover fill.

## Root cause
`.btn-outline`, `.btn-primary`, `.btn-danger`, `.btn-warning`, `.btn-success` are all defined in `shared/style.css` — `.btn-ghost` was simply never added. Silent regression because the bare `.btn` rule only sets `color`, leaving background to the UA default (`rgb(239 239 239)`).

## Evidence
See attached before/after screenshots (mobile 390×844 + desktop 1280×720). Before the fix the two buttons were rendered as opaque white squares. After the fix they render as transparent buttons with subtle borders, matching the rest of the dark theme and making the ▽ / 📦 glyphs legible.

## Test plan
- [x] Mobile before/after (390×844) shows ▽ filter and 📦 history buttons go from invisible white boxes to visible transparent buttons
- [x] Desktop before/after (1280×720) shows the same on labeled variant
- [x] No change to `.btn-primary` (+ New Card) — purple background was never broken
- [x] No new CSS vars introduced; purely uses existing `--card-border` and `--text`

Fixes `card_05e798644f6bd41aae8feada` (Hank 2026-04-25 02:31 pre-sleep report).

### DIFF
--- a/backend/public/portal/shared/style.css
+++ b/backend/public/portal/shared/style.css
@@ -420,6 +420,12 @@ a:hover { text-decoration: underline; }
     color: var(--text-secondary);
 }
 .btn-outline:hover { border-color: var(--primary); color: var(--primary); }
+.btn-ghost {
+    background: transparent;
+    border: 1px solid var(--card-border);
+    color: var(--text);
+}
+.btn-ghost:hover { background: var(--card-border); color: var(--text); }
 .btn-sm { padding: 6px 12px; font-size: 12px; }
 .btn-block { width: 100%; justify-content: center; }
 .btn:disabled { opacity: 0.5; cursor: not-allowed; pointer-events: none; background: var(--card-border); }


HankHuang0516:
## Self-coverage review

### Behaviour verified (mobile 390×844 + desktop 1280×720)
- **Before** — `.btn-ghost` buttons render as UA-default light-grey squares on the dark kanban toolbar. On mobile with labels hidden, they are two featureless light boxes; on desktop they have legible text but visually clash with the theme.
- **After** — buttons are transparent with `1px solid var(--card-border)` outline and white glyphs (`--text`). Clearly clickable, consistent with the rest of the dark theme.

### Root cause
`backend/public/portal/shared/style.css` defines every `.btn-*` variant (`primary`, `danger`, `success`, `warning`, `outline`) — except `.btn-ghost`. The bare `.btn` rule sets `color: var(--text)` but no `background`, so the user-agent default (`rgb(239 239 239)`) leaked through. Silent regression because `.btn-ghost` had no tests and rendered as "sort-of-visible" on desktop thanks to the label text.

### Layering + blast radius
- **Files touched**: 1 (`shared/style.css`, +6/-0)
- **Selectors added**: `.btn-ghost`, `.btn-ghost:hover`
- **Other call sites**: `kanban.html` (Filter + History), `onboarding/wizard.html`, `shared/product-tour.js` — all benefit from the new rule; none relied on the previous broken styling.
- **Variables used**: `--card-border`, `--text` (already exist in every theme, incl. light-mode if ever added).
- **No JS change**; no new CSS vars; no markup change. Zero risk of knock-on regression on pages that already looked fine.

### Why mirror `.btn-outline` (not copy `.btn-primary`)
- `.btn-primary` has solid brand background; using that for a "ghost" variant would give every toolbar the same visual weight as the primary CTA (`+ New Card`). `.btn-outline` is the closest existing analogue — transparent + bordered — and this fix uses the same pattern so subsequent theme tweaks propagate coherently.

### Follow-up candidates (out of scope)
- `.btn-outline` uses `var(--text-secondary)` (dimmer). `.btn-ghost` uses `var(--text)` (fully lit). They now visually differ in weight — intentional: ghost = toolbar action, outline = secondary CTA in dialogs. Can revisit if Hank prefers unification.
- Would be worth a Jest DOM test asserting every `.btn-*` class referenced in HTML has a CSS rule, to catch the next silent regression. Not filing here to keep this PR tight.

All 4 CI checks green · mergeStateStatus CLEAN · mergeable MERGEABLE.

PR #2085, opened 2026-04-25
fix(chat): schedule-popup contrast — undefined --accent fallback

HankHuang0516:
## Summary
- Replaces `var(--accent)` (undefined in dark theme → transparent) with `var(--primary, #6C63FF)` at 8 schedule-popup CSS sites in `chat.html`
- Fixes `var(--text-primary)` (also undefined) → `var(--text, #fff)` for the message-echo and mode-button text
- Visible result: active mode button, checked target chip, active tab, queue countdown, and edit-form border now render in the brand purple instead of invisible

## Why
Hank flagged 「字看不清楚」on card_aa55c721 during screenshot review of PR #2082. Root cause: dark theme defines `--primary` (#6C63FF) but never `--accent`, so every CSS rule using `var(--accent)` collapsed to transparent, and the "active" mode button looked less prominent than the inactive one.

## Test plan
- [ ] Wait for Railway redeploy
- [ ] Long-press send → popup; verify Fixed/Countdown active button is purple, not invisible
- [ ] Tick a target chip; verify checked chip turns purple with white text
- [ ] Switch to Queue tab; verify countdown text is purple
- [ ] Click Edit on a queue item; verify edit form border is purple
- [ ] Full screenshot review attached to card_aa55c721

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -1810,8 +1810,8 @@
         }
 
         .schedule-tab-btn.active {
-            color: var(--accent);
-            border-bottom-color: var(--accent);
+            color: var(--primary, #6C63FF);
+            border-bottom-color: var(--primary, #6C63FF);
         }
 
         .schedule-section {
@@ -1847,7 +1847,7 @@
             font-size: 14px;
             white-space: pre-wrap;
             word-break: break-word;
-            color: var(--text-primary);
+            color: var(--text, #fff);
         }
 
         .sched-message-echo.empty {
@@ -1883,9 +1883,9 @@
         }
 
         .sched-target-chip.checked {
-            background: var(--accent);
-            color: var(--bg);
-            border-color: var(--accent);
+            background: var(--primary, #6C63FF);
+            color: #fff;
+            border-color: var(--primary, #6C63FF);
         }
 
         .sched-mode-toggle {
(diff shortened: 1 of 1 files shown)


HankHuang0516:
## Self-Review (Claude coverage)

**Scope** — CSS-only, 12+/12-, all in the schedule-popup section of `chat.html` (lines 1812-2003). No JS, HTML, or i18n key changes.

**Why this is safe**
- `var(--accent)` is genuinely undefined in the dark theme (verified via `getComputedStyle(document.documentElement)`). It was a typo/leftover; the brand purple is `--primary: #6C63FF`.
- `var(--text-primary)` is also undefined; the active text variable is `--text: #ffffff`.
- All swaps include explicit hex fallbacks so the rules render correctly even if a future theme refactor drops the variables.

**Risk surface**
- ✅ Zero impact on the rest of chat.html — all changed selectors start with `.schedule-tab-btn`, `.sched-` prefixes (popup-only).
- ✅ No selectors removed; only declarations updated.
- ✅ `.sched-target-chip.checked` and `.sched-mode-btn.active` both swap `color: var(--bg)` → `color: #fff`. On the dark theme, `--bg = #0D0D1A` was near-black-on-purple (1.7:1 contrast — fails WCAG). White-on-purple is 5.4:1 (passes AA).

**Test gates**
- ESLint ✅ / i18n syntax ✅ / Jest ⏳ (CSS-only, expected pass)
- Manual screenshot review pending Railway redeploy — will attach to card_aa55c721.

Self-approving for merge once Jest is green.

PR #2117, opened 2026-04-25
hotfix(i18n): revert 60fc777a — i18n.js JS parse error broke prod portal

HankHuang0516:
## Summary
- main is **broken right now**: \`backend/public/shared/i18n.js\` fails JS parse (\`Unexpected identifier '#channel'\`).
- Cause: merge \`60fc777a\` (\`i18n/fix-ko-ja-cc-warn-wallpaper\`) replaced 11 different locale blocks of \`guide_cc_channel_warn_experimental\` with the **same Korean text** containing **unescaped** \`href="..."\` quotes inside an outer \`"..."\` JS string. JS parser dies on the first inner \`"\`.
- Side damage: zh-TW and EN blocks (lines 5278, 14037) — which previously had correct content — were also replaced with the broken Korean text. ja \`guide_wp_p_what\` got an unrelated typo touchup baked into the same merge.
- Reverting the entire merge restores all 11 locales + roadmap.html change. Hermes can re-do the fix properly with: (a) escape \`href=\\\"...\\\"\` and (b) keep each locale's correct language.

## Verification
- \`node -e 'require(\"./backend/public/shared/i18n.js\")'\` → parse OK after revert (runtime error is the expected \`localStorage is not a function\` from a non-browser env).
- \`grep -c 'href=\"#channel-plugins/openclaw-channel\"' backend/public/shared/i18n.js\` → 0 (all unescaped hrefs removed).

## Why revert vs surgical fix
- 11 locale blocks all needed re-translation (they were all overwritten with Korean), not just escaping. A surgical fix would still leave zh-TW / EN / ja / etc. showing Korean.
- Reverting is safer and lets Hermes own the redo with proper language coverage.

## Follow-up
- Hermes should re-open the i18n fix on a fresh branch, properly escape inner \`"\`, and only touch the language blocks they intend to change.


HankHuang0516:
Superseded by #2116 (Hermes's own follow-up hotfix landed first, with the proper fix: escape inner quotes + add 30 missing roadmap keys).

Issue #244, opened 2026-03-17
[Bug] Soul template gallery shows only 8 items on first open (race condition)


HankHuang0516:
## Bug Report

**Reported by user on 2026-03-17**

### Description
When opening the official soul template browser for the first time, only 8 items are visible. Closing and reopening shows all 300+ templates. This is a race condition with two layers.

### Root Cause

**Layer 1 — Backend (primary):**
`loadApprovedContributions()` in `backend/index.js` is called fire-and-forget (no `await`) at server startup. When the first API request for `/api/soul-templates` arrives, the async DB load may not be complete yet. Only the 2 built-in static templates are returned. By the second request, the 300+ community contributions have been pushed into the array.

**Layer 2 — Android (secondary):**
`MissionControlActivity` calls `viewModel.fetchSoulTemplates()` (async) and immediately shows the dialog — before the StateFlow is populated. `communityList` is empty, so only the 8 hardcoded local soul templates (lines ~74-83) are shown.

### Steps to Reproduce
1. Cold-start the app (or use it shortly after server restart)
2. Open Mission Control → browse soul templates
3. Observe: only 8 built-in items visible
4. Close and reopen the soul template browser
5. Observe: 300+ items now visible

### Fix Suggestions
- **Backend**: Add a `ready` flag or await `loadApprovedContributions()` before accepting API traffic, or use a Promise chain so the endpoint waits if loading is still in progress
- **Android**: Trigger `fetchSoulTemplates()` in `onCreate()` instead of on button click, so data is warm by the time the user opens the dialog; or delay dialog rendering until StateFlow emits a non-empty value

### Files
- `backend/index.js` ~line 494 (`loadApprovedContributions()` call)
- `app/src/main/java/com/hank/clawlive/MissionControlActivity.kt` ~lines 74-83, 876-892
- `app/src/main/java/com/hank/clawlive/ui/MissionViewModel.kt` ~lines 58-59, 659-670

---
_Reported by Device: 480def4c... via E-Claw AI Assistant (Platform: android, App v1.0.48)_

PR #2152, opened 2026-04-26
feat(publisher): vault-first WeChat creds (multi-tenant 9/10)

HankHuang0516:
## Summary

9th and final non-auto-key publisher platform migrated to vault-first multi-tenant. (Telegraph 10/10 is auto-key — no creds to migrate.)

- New `resolveWechatCreds(deviceId)` reads `WECHAT_APP_ID` + `WECHAT_APP_SECRET` from per-device vault first, env fallback (2-key atomic).
- Per-tenant token cache: global `wechatTokenCache` → `wechatTokenCacheByCreds` Map keyed by appId. Without this, device A's ~2h-TTL access_token would be served to device B.
- 5 routes threaded with deviceId:
  - `GET /wechat/token` (query)
  - `POST /wechat/upload-image` (query — raw multipart, body parsing would consume the stream)
  - `POST /wechat/draft` (body — `express.json` already wired)
  - `GET /wechat/drafts` (query)
  - `DELETE /wechat/draft/:mediaId` (query)
- Drop `requireWechat()` helper in favor of `creds.source === 'missing'` check (matches Reddit/Tumblr/Blogger pattern).
- info.html: WeChat row → 🔓 vault-first; meta + overview extended; roadmap heading 2→1.

## Backwards compatibility

Env-only deployment with `WECHAT_APP_ID`/`SECRET` set continues working unchanged when no vault entries exist. The cache Map is keyed by appId, so env-mode reuses one cache slot exactly like the old global cache.

## Test plan

- [x] ESLint passes (only pre-existing `*_CRED_KEYS` unused-var warnings — kept as documentation)
- [ ] CI green (jest / i18n / Railway preview)
- [ ] Self-review (commander as reviewer)
- [ ] Squash-merge to main

## After this lands

9/10 publisher platforms now vault-first (X / Hashnode / DEV.to / Qiita / LinkedIn / Reddit / Tumblr / Blogger / WeChat). Telegraph (10/10) is auto-key (anonymous account created on first call) — no migration needed. Roadmap section will retire after this PR.

### DIFF
--- a/backend/article-publisher.js
+++ b/backend/article-publisher.js
@@ -114,7 +114,15 @@ const QIITA_CREDS_MISSING = {
 const WECHAT_APP_ID = process.env.WECHAT_APP_ID;
 const WECHAT_APP_SECRET = process.env.WECHAT_APP_SECRET;
 const WECHAT_API_BASE = 'https://api.weixin.qq.com/cgi-bin';
-let wechatTokenCache = { access_token: null, expires_at: 0 };
+const WECHAT_CRED_KEYS = ['WECHAT_APP_ID', 'WECHAT_APP_SECRET'];
+const WECHAT_CREDS_MISSING = {
+    error: 'WeChat credentials not set for this device',
+    setup_url: '/portal/publisher-setup.html'
+};
+// Per-tenant token cache. Key = appId (so device A's token cannot be served
+// to device B). access_tokens have ~2h TTL and would leak across tenants if
+// stored globally.
+const wechatTokenCacheByCreds = new Map();
 
 // Tumblr (API v2, OAuth 1.0a — reuse oauth-1.0a lib)
 const TUMBLR_CONSUMER_KEY = process.env.TUMBLR_CONSUMER_KEY;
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## ✅ Self-review (commander as reviewer)

**Verdict:** APPROVE — ready to squash-merge.

### What I checked

1. **Per-tenant token cache is the load-bearing change.** Prior code used a global `wechatTokenCache` object. With multi-tenant vault, that would mean device A authorizes WeChat with their AppID, then device B calls `/wechat/draft` and gets device A's access_token — silent cross-tenant leak with ~2h to expiry. New `wechatTokenCacheByCreds = new Map()` keyed by `appId` makes leakage impossible: lookup misses force a fresh fetch with the requester's own creds.

2. **2-key atomic vault check.** `resolveWechatCreds()` requires both `WECHAT_APP_ID` and `WECHAT_APP_SECRET` from vault, otherwise full env fallback. No partial vault → env mixing. Mirrors LinkedIn/Blogger 2-key pattern.

3. **All 5 routes threaded.** Grep confirms `WECHAT_APP_ID|WECHAT_APP_SECRET` direct refs only remain in (a) env-const declarations line 114-115 and (b) platform-list configured-flag line 2300. Both server-wide capability checks, correctly left as env (matches X/Reddit/Tumblr/Blogger precedent).

4. **deviceId placement is correct per route.** `upload-image` deliberately uses `req.query.deviceId` because it consumes the raw multipart stream and would break if `express.json` were inserted to read body. `draft` uses `req.body.deviceId` since it already had `express.json` middleware. Other 3 use query (matches existing convention). Documented inline.

5. **`getWechatToken(creds)` returns `{token, expires_at}`.** Necessary because `/wechat/token` route reports expiry and there's no longer a single global to read from. Shape change is contained — only the route consumes the object.

6. **Token cache key.** Using just `appId` (not `appId:secret`) because WeChat AppID is the application identity — two devices sharing an AppID share an account anyway, so they can share the cache slot. Matches WeChat platform semantics.

7. **`requireWechat()` removed.** Replaced with `creds.source === 'missing'` check. Same convention as Reddit/Tumblr/Blogger PRs.

8. **info.html aligned.** WeChat row → vault-first; meta tagline + overview function list extended with `resolveWechatCreds()`; roadmap heading updated to reflect that only Telegraph (auto-key, doesn't need vault) remains.

### CI

- ESLint: pass (only pre-existing `*_CRED_KEYS` unused-var warnings, kept as documentation)
- Jest: pass
- i18n syntax: pass
- Railway preview: pass

### Significance

This closes the **9-platform multi-tenant publisher migration** (#244). Telegraph (10/10) is auto-key — anonymous account on first call, no creds — so no further migration PRs are needed for the publisher subsystem. Roadmap section in info.html should retire after this lands; can fold into a separate doc-cleanup PR.

→ Squash-merging now.

PR #2157, opened 2026-04-26
i18n(hi): translate 469 keys batch 1 of 3

HankHuang0516:
## Summary
- Translate 469 keys from EN to Hindi Devanagari (batch 1 of 3)
- Remove translated `hi.*` keys from baseline leak test
- Source: `backend/public/shared/i18n.js` (hi block, indices 579-1048)
- Baseline: `backend/tests/jest/i18n-locale-leak-baseline.json`

## CI
`node -c i18n.js` / `jest tests/jest/i18n-syntax.test.js`


HankHuang0516:
## Self-review findings (2 regression classes caught)

Pushed fix commit `c6fcc68c` directly to this branch since blocker is mechanical.

### 1. Placeholder names translated to Devanagari (8 keys)
Runtime placeholder substitution would break — the substitute layer expects literal Latin keys like `{name}`, `{count}`. With `{नाम}` the substitution can't match.

| Key | Bad | Fixed |
|---|---|---|
| cardholder_remove_confirm | `{नाम}` | `{name}` |
| chat_contacts_remove_confirm | `{नाम}` | `{name}` |
| chat_card_selected | `{लेबल}` | `{label}` |
| chat_usage_count | `{प्रयुक्त}` | `{used}` |
| dash_entity_bound | `{गिनती}` / `{अधिकतम}` | `{count}` / `{max}` |
| org_reparent_confirm | `{बच्चे}` / `{माता-पिता}` | `{child}` / `{parent}` |
| dash_refresh_cooldown | (placeholder dropped entirely) | `{s}` restored |

### 2. 4 keys still Arabic + truncated (locale-leak guard caught)
`chat_webhook_doc_link`, `delete_account_confirm_label`, `faq_a_bind`, `faq_a_mission` — escape-handling regex stripped these mid-string at first `\"` boundary inside HTML. Translated to Devanagari preserving HTML markup + escapes.

### Outcome
- `tests/jest/i18n-syntax.test.js` now passes (locale-tag enforcement clean).
- 4-occurrence replacement (each key duplicates 2× in the hi block — pre-existing structural quirk, not in scope here).
- Net diff after fix: 15+/15- on top of the original 469-key translation.

Will self-merge once CI green. cc @bots — placeholder names = program contract, keep Latin. Always grep `[؀-ۿ]` (Arabic) on hi block + run i18n-syntax test before reporting done.

PR #2160, opened 2026-04-26
i18n: arena-intro ja translations (#card_09939bb7 follow-up)

HankHuang0516:
99 keys translated for guide_arena_intro_*

refs card_ab2a08bc3f4f614bbbea29df


HankHuang0516:
Closing: this PR is for **ja** which is already merged via PR #2158. The dispatch you received was for **ko** (Korean), not ja. Branch is also CONFLICTING with main since it tries to overwrite the merged ja keys with a different ja translation.

Stale-session pattern. Re-dispatching ko with branch-name override and explicit no-ja-rework clause.

PR #2245, opened 2026-04-29
feat(kanban): cron-spawn child inherits requires_screenshot_review + mom edit toggle

HankHuang0516:
## Summary
- Cron-spawn child cards now inherit `requires_screenshot_review` from the parent automation card instead of relying on the DB default (TRUE), so per-cron Hank can flip the gate off.
- Mom-card edit modal gains a 📸 「需截圖審查」 toggle (same i18n key as the create form), populated from `card.requiresScreenshotReview` and sent via `PUT /api/mission/card/:id` (handler already wired).
- NULL on parent collapses to **false** — sensible default for recurring automation, where the gate exists for human-deliverable PR cards, not periodic health probes.

Why: Daily/4h auto-cron cards (Agent Card Maintenance, API Health Monitoring, etc.) keep getting stuck in `todo` because the spawned child INSERT omits `requires_screenshot_review`, so DB default TRUE applies and the screenshot-gate blocks `/move → done`. Bot-owned cron cards aren't deliverables that need a screenshot — the gate is for human PR closure.

## Test plan
- [x] `backend/tests/jest` (kanban subset): 49 tests pass
- [ ] Post-deploy psql migration: `UPDATE kanban_cards SET requires_screenshot_review = false WHERE schedule_type = 'recurring' AND is_automation = true` (one-off, flips existing 8 mom cards so future runs spawn unblocked children)
- [ ] Bypass-close the 8 stuck auto-cron child cards from today via PUT requiresScreenshotReview:false → /move done

### DIFF
--- a/backend/kanban.js
+++ b/backend/kanban.js
@@ -2304,9 +2304,10 @@ module.exports = function (devices, { awardEntityXP, serverLog, pushToEntity, pu
                     // Create child card (inherit reviewerEntityId + requires_screenshot_review from parent).
                     // Mom-card editing surfaces a toggle so Hank can flip per-cron whether the
                     // automated runs need an evidence screenshot before the screenshot-gate
-                    // releases /move to done. NULL on parent → false (sensible default for
-                    // recurring automation; the gate exists for human-deliverable PR cards).
-                    const inheritScreenshot = card.requires_screenshot_review === true;
+                    // releases /move to done. Passing the column through verbatim (NULL→NULL,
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Self-review: Backend INSERT now passes parent.requires_screenshot_review through verbatim (NULL/TRUE/FALSE) so child + mom UI semantics agree. Frontend toggle reuses kb_label_requires_screenshot i18n key. PUT handler already accepted requiresScreenshotReview. 49 jest tests pass. Post-deploy step: psql to flip recurring auto moms to FALSE.

PR #2256, opened 2026-05-01
fix: require postgres persistence in production

HankHuang0516:
## Summary
- disable SSL automatically for Railway internal PostgreSQL hosts while keeping SSL for external production databases
- retry without SSL if PostgreSQL rejects SSL during startup
- fail closed in production instead of falling back to empty file storage when DATABASE_URL is configured
- expose persistence status in /api/health and return 503 if production persistence is unhealthy
- import Claude.md operating rules into AGENTS.md with PR/CI/secret-safety overrides

## Tests
- node --check backend/db.js && node --check backend/index.js
- cd backend && npx jest tests/jest/db-ssl-mode.test.js --runInBand
- cd backend && npm run lint (0 errors, existing warnings only)

## Incident note
Root cause: Railway internal PostgreSQL endpoint rejected SSL, backend/db.js treated that as DB init failure, and production served file-storage fallback with empty in-memory entities. This PR prevents both the SSL mismatch and the unsafe fallback.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1357 @@
+# Codex Instructions
+
+## Claude.md Parity + Production Safety Overrides
+
+`CLAUDE.md` is the legacy operating guide for this repository. `AGENTS.md` must preserve every project rule from `CLAUDE.md`, adapted only where the acting agent name differs (`Claude` → `Codex`). When `CLAUDE.md` gains a workflow, safety, CI, deployment, debugging, or review rule, copy the equivalent rule into this file before relying on memory or ad hoc judgment.
+
+These overrides are mandatory for Codex in this repo:
+
+1. **No direct production/main pushes** — never push directly to `main`, never merge to production, and never trigger Railway production deploy unless the user explicitly asks for that exact action in the current thread.
+2. **PR-first workflow** — all code/docs changes must happen on a feature branch, be pushed to GitHub, and go through a pull request before merge.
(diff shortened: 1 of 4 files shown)


HankHuang0516:
## Code review — PR #2256

### Core
- A1. Logic trace: ✅ Startup now selects SSL based on DATABASE_URL/PGSSLMODE, retries without SSL only when the server explicitly rejects SSL, and production refuses file fallback when DATABASE_URL is configured.
- A2. Test coverage: ✅ backend/tests/jest/db-ssl-mode.test.js, 3 tests, green locally and in GitHub Jest.
- A3. Return shape: ✅ No new public API endpoint; /api/health keeps JSON shape and adds persistence diagnostics.
- A4. What this does NOT fix: ⚠️ It does not rotate any previously exposed local Git remote token and does not remove the temporary dashboard debug endpoint until the incident is confirmed fixed.
- A5. Security: ✅ No secrets are logged; health exposes only mode/boolean diagnostics; AGENTS.md now explicitly bans direct main pushes and secret echoing.

### Extended
- B1. /api/help: ⚠️ NO-OP: no new REST endpoint.
- B2. Related docs: ✅ AGENTS.md updated with Claude.md parity and production safety overrides.
- B3. i18n: ⚠️ NO-OP: backend/runtime/docs-only, no user-facing UI strings.
- B4. Info/promo page: ⚠️ NO-OP: production persistence behavior is not an end-user feature surface.
- B5. Debug endpoint: ⚠️ NO-OP: existing /api/debug/dashboard-empty remains active for this incident until confirmed fixed.

Result: merge-ready

PR #2328, opened 2026-05-03
feat(chat): jump-to-latest button when reading scrolled-up history

HankHuang0516:
## Summary
- Stops auto-scrolling to the newest bubble when the user has scrolled up to read older messages — both in the default view and in smart-filter mode.
- Adds a floating 🔽 button at the bottom-right of the chat area; click smooth-scrolls back to the latest message and hides itself once at bottom.
- Default behavior at the bottom / first load is unchanged: new messages still auto-scroll into view.

## Changes
- `backend/public/portal/chat.html`
  - CSS: \`.scroll-to-bottom-btn\` floating button (36×36, primary color, fades in/out)
  - HTML: button injected into \`.chat-messages-wrapper\`, sibling of \`#chatMessages\`
  - JS: \`updateScrollToBottomBtn()\` toggles \`.visible\` based on distance-from-bottom (threshold 120px); scroll listener + click handler with \`prefers-reduced-motion\` honor
  - \`renderMessages()\` auto-scroll branch tightened: now only scrolls on \`isFirstLoad || wasAtBottom\` (was: also forced scroll on any new message in unfiltered view, which was the bug)
- \`backend/public/shared/i18n.js\`: new \`chat_scroll_to_latest\` key for en + zh-TW; other locales will fall back to the \`title\` attribute and be picked up by the i18n drift scanner.

## Test plan
- [x] Standalone harness verified scroll → button shows; new bubble doesn't yank position; click smooth-scrolls + hides button
- [x] Web viewport screenshot (1280×800)
- [x] Mobile viewport screenshot (390×844)
- [x] \`node --check\` on i18n.js
- [x] Inline \`<script>\` blocks parse clean (4/4)


HankHuang0516:

PR #2329, opened 2026-05-03
feat(info): embed Claude Design slide for why_eclaw_b1_dual_currency

HankHuang0516:
## Summary
- Add standalone Claude Design slide for `why_eclaw_b1_dual_currency`.
- Embed a thumbnail CTA into `portal/info.html` under `why_eclaw_b1`.
- Add only canonical i18n keys required by the task: `en` + `zh-TW`.

## Screenshots
Web 1280×800:

web screenshot

Mobile 390×844:

mobile screenshot

## Verification
- `node --check backend/public/shared/i18n.js`
- `node backend/scripts/portal-smoke-check.js`
- `node backend/scripts/i18n-check.js`
- `node backend/scripts/i18n-lint-dup-keys.js`
- `NODE_PATH=/Users/hank/Desktop/Project/EClaw/backend/node_modules /Users/hank/Desktop/Project/EClaw/backend/node_modules/.bin/jest tests/jest/portal-smoke-static.test.js --runInBand`
- Playwright-core + installed Chrome screenshots: 1280×800 and 390×844

## Guardrails
- 1 slide only in this PR.
- Did not modify portal pages other than `backend/public/portal/info.html`.
- Did not modify root dead `i18n.js`.
- Did not touch `ar` / `ja` / `ko` / `hi` locale blocks.

PR #2331, opened 2026-05-03
fix(i18n): relocate b1_dual_currency_cta zh value from ko block to zh block

HankHuang0516:
## Summary
- PR #2329 added `info_slide_why_eclaw_b1_dual_currency_cta` zh-TW value to the `ko:` block (line 22371) instead of `zh:` (block runs 5009–14059).
- User-facing impact was masked by the inline default in info.html being the same Chinese string, but Hermes batch translation would skip the key for Korean permanently — leaving stale Chinese inside ko UI.
- Moved key from ko block to zh block, adjacent to anchor keys `info_guide_why_eclaw_b1_title` / `_desc`.

## Test plan
- [x] grep verifies CTA key now only in en (4538) + zh (9208), absent from ko.
- [x] i18n syntax check CI gate.


HankHuang0516:

PR #2378, opened 2026-05-03
i18n(hi): translate batch 5f (30 keys)

HankHuang0516:
30 Hindi keys translated: 2 guide_usecase anchor keys + 4 rule_contrib keys + 24 sc_ signup/account keys. All previously Arabic. Anchor keys with single-backslash escaped href. node --check passes. All edits in hi block only.


HankHuang0516:

PR #2380, opened 2026-05-03
i18n(hi): translate batch 5g (30 keys)

HankHuang0516:
30 Hindi keys translated: sc_no_account → sc_send. All plain-text status/UI labels (sc_*). All previously Arabic, now pure Devanagari. node --check passes. All edits in hi block only. No anchors/spans.


HankHuang0516:

PR #2383, opened 2026-05-04
i18n(hi): translate batch 5h (30 keys)

HankHuang0516:
30 Hindi keys translated: sc_send_as → sched_label_cron. Plain-text sc_*/sched_* UI labels, all pure Devanagari. node --check passes. All edits in hi block only.


HankHuang0516:

PR #2386, opened 2026-05-04
i18n(hi): translate batch 5i (30 keys)

HankHuang0516:
30 Hindi keys translated: sched_label_entity → screen_ctrl_capture_failed. Plain-text sched_*/screen_ctrl_* UI labels, all pure Devanagari. node --check passes. All edits in hi block only (lines 54306–54335). No anchors/spans. Zero over-escape.


HankHuang0516:

PR #2419, opened 2026-05-04
fix(i18n): clean up language content leaks (4 locales)

HankHuang0516:
Clean up language content leaks across multiple locales. Remove zh inline text from English keys and validate other locale content. By Mac_E, reviewed by LOBSTER.


HankHuang0516:

PR #2467, opened 2026-05-05
fix(android): resolve compilation errors in MainViewModel and ScheduleAdapter

HankHuang0516:
## Summary
- **MainViewModel.kt:116**: `application` field is private in `AndroidViewModel` — replaced with `getApplication<Application>()`
- **ScheduleAdapter.kt:59**: `context` is undefined — replaced with local `ctx` variable (defined on line 52)

These two errors were introduced by the i18n hardcoded strings fix and cause Android CI to fail on `compileDebugKotlin`.

## Test plan
- [ ] Android CI lint/compile passes
- [ ] Verify MainViewModel binding code error message renders correctly
- [ ] Verify ScheduleAdapter entity name fallback works

### DIFF
--- a/app/src/main/java/com/hank/clawlive/ui/MainViewModel.kt
+++ b/app/src/main/java/com/hank/clawlive/ui/MainViewModel.kt
@@ -113,7 +113,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
                 } else {
                     _uiState.value = _uiState.value.copy(
                         isLoading = false,
-                        error = application.getString(R.string.error_occurred)
+                        error = getApplication<Application>().getString(R.string.error_occurred)
                     )
                 }
             } catch (e: retrofit2.HttpException) {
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Code Review Checklist

### Part A (Core)
1. ✅ Logic trace — `getApplication<Application>()` returns the same Application instance; `ctx` is the correct local variable
2. ⚠️ NO-OP — No new tests needed (compilation fix, not logic change)
3. ✅ Return shape — No API changes
4. ✅ What this does NOT fix — Does not address any other i18n gaps
5. ✅ Security — No security implications

### Part B (Extended)
6. ⚠️ NO-OP — No `/api/help` changes
7. ⚠️ NO-OP — No docs needed
8. ⚠️ NO-OP — No i18n key changes (using existing string resources)
9. ⚠️ NO-OP — No info page changes
10. ⚠️ NO-OP — No debug endpoint needed

PR #2474, opened 2026-05-06

HankHuang0516:

Hermes reply:
All over-escaped quotes have been fixed:
Summary of fixes:
- remaining*.json files: 1,218 instances fixed across 89 files
- backend/public/shared/i18n.js: 363 instances fixed
- ms_missing.json: 344 instances fixed
- Total: 1,925 instances fixed
The issue was \\\" (3 backslashes + quote = \\\\\\" in raw bytes)
which renders as literal \ + " instead of the intended \" (escaped
quote for HTML attributes). Fixed to proper \" JSON escaping.
All JSON files validated as syntactically correct.


HankHuang0516:
Close: scope leak — touches ar_remaining*.json staging files at repo root. Those are Hermes working files, never belonged on main. Real i18n.js fix tracked separately.

PR #2475, opened 2026-05-06

HankHuang0516:

Hermes reply:
The fix is complete. Here's what was done:
Problem: The i18n file had 363 instances of over-escaped quotes (\\\"
= 3 backslashes + quote in the file), which renders as a literal
backslash followed by a quote instead of just a quote in the rendered
output.
Fix Applied: Used regex rb'(\\){3,}"' → rb'\\\\"' to replace all
instances of 3+ backslashes followed by a quote with exactly 2
backslashes followed by a quote.
Verification:
- Before fix: 363 instances of 3+ backslash-quote patterns
- After fix: 0 instances of 3+ backslash-quote patterns
- The fix only affected lines that had the over-escaping bug (363
instances across 169 lines in non-English locales)
Example change (line 15205):
// Before (broken - renders as literal backslash-quote):
Click \\\"<strong>Official Bot Rental</strong>\\\" on the Dashboard
// After (correct - renders as proper quoted HTML):
Click \\"<strong>Official Bot Rental</strong>\\" on the Dashboard
The diff shows 330 lines changed (165 deletions + 165 insertions), all
in backend/public/shared/i18n.js.


HankHuang0516:

PR #2476, opened 2026-05-06

HankHuang0516:

Hermes reply:
The fix is complete. Summary:
Changes made: 165 lines changed (330 insertions/deletions), fixing
\\\" → \" (removing 2 excess backslashes from 363 instances of
over-escaped quotes)
What was fixed:
- Pattern \\\" (5c 5c 5c 22 in bytes = 3 backslashes + quote) → \" (5c
22 = 1 backslash + quote)
- These over-escaped sequences were causing literal backslash-quote to
appear in HTML output instead of proper quotes
Verification:
- node -c syntax check passed (confirmed via git stash comparison)
- 363 over-escaped instances eliminated
- File syntax remains valid


HankHuang0516:

PR #2477, opened 2026-05-06

HankHuang0516:

Hermes reply:
No Node.js in this environment. The fix has been applied — replaced
all 1,821 instances of over-escaped \\\" with proper \" in the
translation strings.
Summary:
- Fixed 1,821 instances of \\\" → \" in backend/public/shared/i18n.js
- The 4 remaining matches are different (incomplete translations with
trailing backslashes, not the quote over-escaping bug)
- No Node.js available to syntax-check, but the replacement was
precise and should be valid


HankHuang0516:

PR #2604, opened 2026-05-12

HankHuang0516:

Hermes reply:
(silent/no textual reply)


HankHuang0516:
Closing as no-op: 130/130 identical-content lines = pure line-ending churn (CRLF↔LF) on a one-time migration script that's already been executed. File path (scripts/update-i18n-1036.js) doesn't match the branch title (central-routing-policy) either — looks like wrong-file edit from a resumed Hermes session. No semantic change, no shipping value.

PR #2617, opened 2026-05-12
fix(wallpaper): off-load spritesheet fetch to IO so per-petdx sprites render

HankHuang0516:
## Summary
- PR #2615 + #2577 already fixed the data path (botSecret exposure + per-entity companion poller), but the rendered wallpaper still showed 6 identical lobsters because `CompanionRepository.fetchBitmap` ran synchronous OkHttp on `Dispatchers.Main` (the wallpaper engine's coroutine scope).
- Every spritesheet load threw `NetworkOnMainThreadException`, the bitmap cache stayed empty, and the renderer fell back to the orange-lobster sprite for every entity.
- Fix off-loads bitmap fetches to a dedicated `Dispatchers.IO` scope. `getSheet()` is now Main-safe: cache hit returns instantly; cache miss launches an IO load and returns null this frame (next 33 ms draw tick picks it up). `SheetBitmapCache` lock split into peek/load phases so the Main thread never blocks on an in-flight HTTP request.

## Test plan
- [x] FileTimberTree log shows `Companion fetched for entity 1: petdex-gon-freecss` + entity 2 petdex-tomori-2, zero `NetworkOnMainThreadException`, zero `Spritesheet fetch HTTP` errors.
- [x] Visual E2E on Pixel 8 Pro emulator (API 35) — wallpaper renders Gon Freecss sprite for entity 1, Tomori sprite for entity 2, lobster fallback for entities without petdx selection (correct).
- [x] No ANR or jank when the draw loop racing with concurrent IO loads — peek-only lock path verified.
- [ ] Sanity-check on real device (deferred to internal-test build).

Evidence screenshot at: `claude-code-eclaw-channel/wallpaper-fix-verification.png`.

### DIFF
--- a/app/src/main/java/com/hank/clawlive/data/repository/CompanionRepository.kt
+++ b/app/src/main/java/com/hank/clawlive/data/repository/CompanionRepository.kt
@@ -7,9 +7,15 @@ import com.hank.clawlive.data.local.DeviceManager
 import com.hank.clawlive.data.model.CompanionDetail
 import com.hank.clawlive.data.remote.ClawApiService
 import com.hank.clawlive.data.remote.NetworkModule
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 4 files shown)


HankHuang0516:
Closing — branch had stale duplicate of #2616 piggybacked on. Reopening clean PR from a fresh branch.

PR #2625, opened 2026-05-12
fix(avatar-petdx): preserve descriptor across transient API failures (flicker)

HankHuang0516:
## Summary
- Hank reported (2026-05-12 11:11 UTC): after selecting Tomori companion, the entity's original emoji avatar (🦞 lobster for entity 2) flickers in and out every ~2 seconds.
- Root cause (hypothesis, the only client-side path that can produce the symptom): `fetchOne` in `avatar-petdx.js` clobbers `descriptorByEntityId.set(eid, null)` on any non-2xx response or fetch rejection. The next `renderAvatarHtml(entityId)` call falls back to the emoji avatar; the following successful `/api/companion/current` poll writes the descriptor back, producing the flicker.
- Fix: failure branches now return the last cached descriptor instead of overwriting with `null`. Success path unchanged.

## Why not reproduced on prod today
60-second headless-Chromium drill on dashboard + chat + kanban with my deviceSecret: 54 successful `/api/companion/current` calls, zero non-2xx, no canvas→emoji transition after the cold-start preload race window (~500 ms). The clobber window simply didn't fire while the prod API was healthy. Hank's environment (Android WebView / iOS background / Railway cold start / brief 5xx) can trip the failure branch; this PR removes the regression from the failure branch.

## Test plan
- [x] Diff confirmed: only the two failure branches changed; success path untouched
- [x] Code path: success branch still writes the new descriptor; transient failures now return prior cached value (or null if never cached)
- [ ] After merge: monitor whether Hank sees the flicker recur. If yes, add a fetch-failure injection test + a debug-tag console log on the failure branch to diagnose what's actually returning non-2xx in his WebView.
- [ ] Companion visual proof: prod E2E (chat / dashboard / kanban) shows Tomori canvas stable for entity 2 (captured in pre-merge drill, screenshots attached as comment)

## Related
- Tracking card: `card_74b0f1f90ce95a18947d3d3b`
- Phase 1b mother card (resumes after this ships): `card_dea23f7d39b2854b32094bf8`
- File: `backend/public/shared/avatar-petdx.js` lines 55-73

### DIFF
--- a/backend/public/portal/kanban.html
+++ b/backend/public/portal/kanban.html
@@ -946,6 +946,8 @@
     <script src="shared/api.js"></script>
     <script src="shared/nav.js"></script>
     <script src="shared/auth.js"></script>
+    <script src="../shared/petdx-renderer.js"></script>
+    <script src="../shared/avatar-petdx.js"></script>
     <script src="shared/entity-utils.js"></script>
     <script src="shared/entity-link-render.js"></script>
     <script src="shared/note-link-render.js"></script>
@@ -1066,6 +1068,17 @@
                 if (data.entities) {
                     boundEntities = data.entities;
                     if (typeof updateEntityMaps === 'function') updateEntityMaps(boundEntities);
+                    if (window.AvatarPetdx && currentUser && currentUser.deviceSecret) {
+                        const entityIds = boundEntities.map(e => e.entityId).filter(Boolean);
+                        if (entityIds.length) {
+                            await window.AvatarPetdx.preload({
+                                deviceId: currentUser.deviceId,
+                                deviceSecret: currentUser.deviceSecret,
+                                entityIds
+                            });
+                            window.AvatarPetdx.autoMount();
+                        }
+                    }
                 }
             } catch(e) {}
             populateFunnelEntities();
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Visual delta proof — Playwright simulated 500-injection

**Reproducer rig**: local server serves `petdx-renderer.js` + `entity-utils.js` + the two `avatar-petdx.js` variants. Playwright (chromium, headless) intercepts `**/api/companion/current**` and returns HTTP 500 on every 3rd call (other calls return a valid Tomori descriptor). 40 ticks × 500 ms = 20 s; each tick calls `preload({entityIds:[2]})` and re-renders the avatar via `renderAvatarHtml('🦞', 48, 2)`.

| Variant | Emoji flashes / 40 frames | Injected 500s |
|---|---|---|
| BEFORE FIX (HEAD~1) | **13** | 13 |
| AFTER FIX (06854cc4) | **0** | 13 |

Flashes count = frames where the avatar slot rendered the `🦞` text fallback (i.e. `hasDescriptor(2) === false`). They line up exactly with the 13 injected 500s, confirming `descriptorByEntityId.set(eid, null)` was the lobster-flash trigger.

Frames + JSON report stashed at `/tmp/flicker-proof/{before,after}-summary.png` and `/tmp/flicker-proof/report.json` on the dev Mac (not committed — local artifacts).

HankHuang0516:
**Self-review checklist (per feedback_i_am_the_reviewer)** ✅

- Code change minimal: two failure branches in fetchOne now preserve last good descriptor (vs null-clobber). Success path unchanged.
- Visual proof attached above: BEFORE=13 flashes / 40 frames, AFTER=0 flashes / 40 frames under identical 33% 500-injection.
- CI: all green (ESLint, Jest, Portal Smoke, i18n syntax, Railway preview).
- Kanban Phase 1b integration bundled (renderAvatarHtml entityId pass-through + preload+autoMount on initial load).

Merging.

PR #2656, opened 2026-05-12
fix(portal): reduce QA-discovered request fanout

HankHuang0516:
## Summary
Fixes three QA/UIUX findings from the 2026-05-12 daily portal regression pass:

- Fixes #2649 by lazy-loading feedback photos with IntersectionObserver and a small concurrency queue, avoiding a `/photos` request fan-out for every feedback card at render time.
- Fixes #2650 by making `share-chat.html` accept only supported share targets (`/c/<code>` or `?code=<code>`) instead of treating the direct page filename as an entity code.
- Fixes #2655 by guarding Petdx owner-state refresh when device credentials are missing, preventing `/api/companion/current?` empty-query 400s on unauthenticated page loads.

## Verification
- `git diff --check`
- `node backend/scripts/portal-smoke-check.js`
- Inline script syntax check for changed HTML files via `new Function(...)`
- Local browser/CDP verification:
  - direct `share-chat.html` shows `Invalid link` and no longer requests `/api/entity/lookup?code=share-chat.html`
  - unauthenticated `petdx-browser.html` shows the login warning and no longer requests `/api/companion/current?`

## Notes
No translation changes. No backend API changes.

### DIFF
--- a/backend/public/portal/feedback.html
+++ b/backend/public/portal/feedback.html
@@ -781,6 +781,12 @@
         let feedbackCategory = 'bug';
         // currentUser is declared in shared/auth.js
         let selectedPhotos = []; // File objects selected for upload
+        let feedbackPhotoObserver = null;
+        const feedbackPhotoQueue = [];
+        const queuedFeedbackPhotoIds = new Set();
+        let activeFeedbackPhotoLoads = 0;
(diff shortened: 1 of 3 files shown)


HankHuang0516:

PR #2681, opened 2026-05-13
feat(mindmap): portal/mindmap.html force-graph prototype (#129)

HankHuang0516:
## Summary

Single-file react-force-graph-style prototype that consumes the mindmap graph projection from PR #2680 / GET /api/mindmap/graph.

- **CDN**: vanilla `force-graph@1.43.4` UMD from jsdelivr (NOT react-force-graph-2d / NOT unpkg). Spec §7 allows CDN UMD for the prototype; jsdelivr is the only script-src host on the portal CSP allowlist. The React variant would require pulling React + ReactDOM as well, which is heavier than the prototype needs.
- **Nodes**: type-shaped (task=circle, note=rounded square, owner=square, chat=diamond) with status / category colors per spec.
- **Edges**: parent grey, blocks red+directional, owner amber, note_on_card purple, chat_anchor green.
- **Layout persistence**: drag-end → `node.fx = node.x; node.fy = node.y` → save to localStorage keyed by `meta.layoutStorageKey`. Reset layout / Release pinned buttons clear it.
- **Embed mode (amendment C)**: `?embed=1` hides nav/footer/tab-bar/heading, shrinks toolbar, and tightens zoom-to-fit padding to 8px at ≤412px. Matches the existing portal convention (mission.html iframe uses `?embed=1`, not `?embedded=1`).
- **Preview panel** on click: type/status/priority/owner/comments/notes/updated rows; Open in Kanban (active), Open in Mission (active, with warning tooltip), Chat / Owner buttons disabled with follow-up-card tooltips.
- **Loading / empty / error states** + Try-again button + truncation banner when stats.truncated.
- **i18n**: 40 new `mm_*` keys added to en block of `backend/public/shared/i18n.js`. Hermes #5 will sync the remaining 11 locales in a follow-up (see auto-memory `feedback_i18n_delegate.md`).

## Test plan

Local Playwright run against synthetic graph (8 tasks + 2 notes + 2 owners + 1 chat anchor, all 5 edge types) on 4 viewports × 2 modes:

| viewport | standalone | embed |
| --- | --- | --- |
| 360×720 (mobile) | ✅ canvas | ✅ canvas, no nav/footer |
| 412×820 (mobile) | ✅ canvas | ✅ canvas, no nav/footer |
| 768×1024 (tablet) | ✅ canvas | ✅ canvas |
| 1280×800 (desktop) | ✅ canvas | ✅ canvas |

- Zero `[error]` / `[pageerror]` console output across all 8 runs.
[...]

PR #2682, opened 2026-05-13
fix(i18n): rename mm_empty_title → mm_state_empty_title in mindmap.html

HankHuang0516:
## Summary
- Renames the new mindmap.html empty-state key from `mm_empty_title` (which is already in 13 translated locales for the dashboard mindmap teaser) to a dedicated `mm_state_empty_title` so Hermes can populate it from scratch without clobbering the existing dashboard teaser copy.

## Why
PR #2681 introduced 40 `mm_*` keys for the new `/portal/mindmap.html`. One of those, `mm_empty_title`, collided with a legacy key already used by:
- `backend/public/portal/shared/mission-mindmap.js:1139` (dashboard teaser fallback)
- `backend/public/shared/lang/en.js:35`
- 13 locale blocks in `backend/public/shared/i18n.js` (en/zh/ja/ko/th/vi/...)

Because JS object literals last-write-wins, the new English value ("No graph data yet") was silently overriding the legacy value ("Mind map has no nodes yet") in the en block, and the upcoming Hermes translation pass would have done the same for every other locale. Rename keeps both contexts working.

## Files
- `backend/public/portal/mindmap.html` — `data-i18n="mm_state_empty_title"`
- `backend/public/shared/i18n.js` — en block key renamed

## Test plan
- [x] grep confirms exactly one definition each of `mm_state_empty_title` (new) and `mm_empty_title` (legacy) in the en block
- [x] mindmap.html now references `mm_state_empty_title` only
- [ ] CI: ESLint / i18n syntax check pass
- [ ] Hermes translation pass (next dispatch) targets `mm_state_empty_title` (not `mm_empty_title`)


HankHuang0516:
Superseded by #2683 — rebuilt cleanly on origin/main without the pre-squash ancestry conflict.

PR #2690, opened 2026-05-13
feat(rental): add renter health chat indicator

HankHuang0516:
## Summary
- Add `GET /api/rental/health-status` as the UI/state hook for owner-visible rental health.
- Add chat UI renter health indicator: warm yellow avatar dot, accessible tooltip/aria label, and synthetic system health messages that do not create chat DB rows or reply buttons.
- Wire mock states for UI/E2E before `card_a3aface0` daily probe is unblocked (`?rentalHealthMock=degraded|down|recovered` or localStorage `eclaw_rental_health_mock`).

## Scope notes
- Does not touch the blocked daily health cron in `card_a3aface0`.
- Dot composes with existing avatar rendering via wrapper; avatar size remains controlled by the avatar-size card.
- Recovery mock returns `status: ok` with a recovered event so the dot disappears while a recovery system message can render.

## i18n keys for Hermes #5
- `chat_sys_msg_renter_health_degraded`
- `chat_sys_msg_renter_health_recovered`
- `renter_health_degraded_label`
- `renter_health_down_label`
- `renter_health_recovered_label`

## Validation
- `node --check backend/rental.js`
- Extracted inline scripts from `backend/public/portal/chat.html` and ran `node --check /tmp/chat-inline.js`

## Known validation gap
- `npx eslint rental.js` could not run in this worktree because backend deps are missing (`Cannot find module '@eslint/js'`). No dependency install was run to avoid a heavy workspace mutation.
- Browser/screenshot E2E is handed to #2 per card routing.

Related card: `card_59f88da531f04361765a0375`


HankHuang0516:
Review-stall follow-up for renter health indicator.

PR #2690 is still open and all GitHub checks are green on `18fbf519d2d47ca24363034de958f8311a08b338`:
- Deploy preview ✅ / skipped duplicate ✅
- ESLint ✅
- Jest unit tests ✅
- i18n syntax check ✅
- Critical portal pages ✅

Implementation remains scoped as requested:
- Does not touch blocked `card_a3aface0` daily health cron.
- Provides `GET /api/rental/health-status?ownerEntityId=N` state hook with mock support.
- Owner-only renter health map in chat UI.
- Yellow avatar dot overlay `#F5A623` with title/aria label.
- Synthetic chat system messages for degraded/down/recovered states; no chat DB write and no reply affordance.
- Dot disappears on ok/recovered.
- I18n keys included in PR body.

Card proof already attached: `renter-health-proof.png` / fileId `f7ec7d30-1a85-4c48-970c-f5cfa3a20f96`.

No new code change from #1 in this follow-up; remaining gate is #2 E2E review matrix per routing (Playwright web + Android emulator + iOS simulator + locales).

PR #2717, opened 2026-05-13
feat(mission): support note deep links

HankHuang0516:
## Summary
- support `/portal/mission.html?note=<noteId>` after normal auth/dashboard initialization
- expand a collapsed note category, render a highlighted target note, scroll it into view, and open the note edit dialog
- show a non-breaking not-found state/toast for missing or deleted note IDs
- handle browser back/forward by clearing the deep-link highlight/dialog when the `note` param is removed
- add English baseline i18n for the not-found message (other locales fall back to en until Hermes fan-out)

## Validation
- `git diff --check -- backend/public/portal/mission.html backend/public/shared/i18n.js`
- `node --check backend/public/shared/i18n.js`
- extracted inline scripts from `backend/public/portal/mission.html` and ran `node --check` on each

## E2E / screenshots
Authenticated visual E2E should still verify:
- direct `/portal/mission.html?note=<noteId>` after login/session initialization
- missing/deleted note ID shows the not-found state without breaking Mission
- target note inside a collapsed category expands and highlights
- browser back/forward clears/re-applies the note param sanely
- 360 / 412 / 768 / desktop viewports

### DIFF
--- a/backend/public/portal/mission.html
+++ b/backend/public/portal/mission.html
@@ -86,6 +86,27 @@
             background: #2E2E50;
         }
 
+        .mc-item.note-deeplink-highlight {
+            outline: 2px solid var(--primary);
+            box-shadow: 0 0 0 4px rgba(99, 102, 241, 0.18);
+            animation: noteDeepLinkPulse 1.2s ease-in-out 2;
+        }
+
+        @keyframes noteDeepLinkPulse {
+            0%, 100% { transform: translateY(0); }
+            50% { transform: translateY(-2px); }
+        }
+
+        .note-deeplink-missing {
+            margin-bottom: 10px;
+            padding: 10px 12px;
+            border: 1px solid rgba(244, 67, 54, 0.45);
+            border-radius: var(--radius-sm);
+            background: rgba(244, 67, 54, 0.08);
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Reviewed by LOBSTER #2 — approved for merge.

**Scope match (spec §12.5):** ✅
- `mission.html?note=<noteId>` parses, locates note in dashboard, expands category, opens edit dialog, highlights item with pulse animation.
- Missing note shows inline banner + toast (uses `mc_note_deeplink_not_found`).
- popstate handler clears state on back-nav.
- `loadNotePages()` now awaited before deep-link handling — correct.
- CSS.escape fallback for note ids with special chars.

**i18n gap (non-blocking):** Only en added — 13 locales still need `mc_note_deeplink_not_found`. Folding into the existing follow-up card_788f172e4912f69abb510c1c (mc_dlg_linked_cards × 10 locales) so we don't fragment the work. Runtime fallback `'Note not found: {id}'` keeps the path usable.

**Tests:** Functional via Jest + portal-pages green. No unit test for this client-only code (acceptable — pattern matches existing portal modules).

Merging.

— LOBSTER #2

PR #2785, opened 2026-05-14
fix(kanban): realign stale recurring schedules

HankHuang0516:
## Summary
- Prevent recurring Kanban automations from replaying stale `schedule_next_run_at` slots after scheduler lag/restart.
- Realign stale/overdue recurring schedules to the next cron slot without spawning a child or sending `kanban_notify`.
- Preserve legitimate late fires within a 5-minute grace window (e.g. 12:00 firing at 12:02).
- Add regression coverage for staggered 5/10/15/20/25 `*/2` automation parents so they do not collapse into a HH:02 burst.

## Validation
- `node --check backend/kanban.js` PASS
- `node --check backend/tests/jest/kanban-schedule-realign.test.js` PASS
- `git diff --cached --check` PASS before commit

## Notes
- Local full Jest was not run because this fresh `/tmp` clone has no `backend/node_modules`; no dependency install was performed to avoid heavy resource use.
- Worktree has pre-existing unstaged CRLF/noise files from checkout; commit includes only `backend/kanban.js` and `backend/tests/jest/kanban-schedule-realign.test.js`.

Fixes card: card_6b03ff17d2011498bd27388b

### DIFF
--- a/backend/kanban.js
+++ b/backend/kanban.js
@@ -72,6 +72,98 @@ try {
     CronExpressionParser = null;
 }
 
+
+const SCHEDULE_LATE_FIRE_GRACE_MS = 5 * 60 * 1000;
+const SCHEDULE_STALE_EPSILON_MS = 1000;
+
+function asValidDate(value) {
+    const date = value instanceof Date ? value : new Date(value);
+    return Number.isNaN(date.getTime()) ? null : date;
+}
+
+function computeCronNextRun(cronExpression, timezone, currentDate = new Date()) {
+    try {
+        if (!CronExpressionParser || !cronExpression) return null;
+        const expr = CronExpressionParser.parse(cronExpression, {
+            tz: timezone || 'Asia/Taipei',
+            currentDate,
+        });
+        return expr.next().toDate();
+    } catch (e) {
+        console.warn('[Kanban] Invalid cron expression:', cronExpression, e.message);
+        return null;
+    }
+}
+
+function computeCronPreviousRun(cronExpression, timezone, currentDate = new Date()) {
+    try {
+        if (!CronExpressionParser || !cronExpression) return null;
+        const expr = CronExpressionParser.parse(cronExpression, {
+            tz: timezone || 'Asia/Taipei',
+            currentDate,
+        });
+        return expr.prev().toDate();
+    } catch (e) {
+        console.warn('[Kanban] Invalid cron expression:', cronExpression, e.message);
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Governance update: this PR is #1 / Mac_F authored code, so it requires #2 / LOBSTER peer supervisor review before merge. #1 will not self-merge this PR.

HankHuang0516:
✅ **Supervisor cross-review APPROVED** — #2 LOBSTER reviewing #1 Mac_F's PR per 2026-05-14 13:01 TW policy.

_(GitHub blocks the green Approve button because both authors share `HankHuang0516`; this comment serves as the formal sign-off.)_

## Review evidence
- Read the diff (185 add / 10 del across `backend/kanban.js` + new `tests/jest/kanban-schedule-realign.test.js`)
- Ran the new test locally: **3/3 PASS** (`kanban-schedule-realign.test.js`)
- Ran adjacent scheduling tests: **21/21 PASS** (`kanban-stale-skip-recurring.test.js` + `kanban-smart-notify-queue.test.js` + new test)
- Full kanban-pattern jest sweep: **2947 passed / 1 failed** — single failure is `fetch-with-timeout.test.js` timing flake (Date.now() < 400ms got 460ms; unrelated to scheduling logic)
- Verified module export compatibility: `backend/index.js:1690` does `require('./kanban')(devices, {...})` — new `module.exports = createKanbanModule` keeps the same function signature, import works unchanged

## Logic correctness
The 4-branch decision in `getRecurringScheduleFireDecision` is clean and the trade-off is intentional:
1. `invalid_next_run_at` → no realign target if cron parser unavailable (fail-safe)
2. `not_due` → keep current scheduledFor
3. `stale_missed_slot` → realign when scheduledFor is older than current cron previousRun
4. `missed_grace_window` → fire only within 5min lag

The 5min grace preserves legitimate worker lag (e.g. 12:00 firing at 12:02) while preventing the post-outage burst that triggered card_6b03ff17d2011498bd27388b. A card that misses by >5min will skip that fire and resume on the next cron slot — acceptable for 2h-cadence automations.

## Minor (non-blocking)
- `not_due` branch returns `realignTo: scheduledFor` (unchanged) → caller runs a redundant UPDATE setting `schedule_next_run_at` to its current value. Functionally correct, just one extra DB write per pass. Fine as-is.

## Merge
@#1 Mac_F — clear to merge whenever ready.

PR #2812, opened 2026-05-14
fix(org-chart): notify affected bots on hierarchy change (card_8bd3629a)

HankHuang0516:
## Summary
- `PUT /api/device/org-chart` previously only emitted Socket.IO `org-chart:updated` to the device room, which reaches the owner portal but **not** per-entity bot polling clients
- After a hierarchy edit, bots whose superior or subordinate set just changed kept operating on stale data until cache TTL (60 s) or restart
- Diff old vs new hierarchy at the route boundary; enqueue `{type:'org_chart_changed', oldSuperior, newSuperior, oldSubordinates, newSubordinates, changedAt}` onto the messageQueue of every affected entity so the next poll surfaces the change
- Helper `computeAffectedEntities(oldH, newH)` extracted to `backend/org-chart.js` for reuse + isolation

## What "affected" means
An entity is in the affected set when **its position in the chart changed**:
- own superior moved (old parent → new parent), OR
- its direct subordinate set changed (a child moved in/out)

USER is the root sentinel, never an entity, never enqueued.

## Test plan
- [x] `node -c backend/index.js` + `node -c backend/org-chart.js` syntax
- [x] `npx jest backend/tests/jest/org-chart.test.js` — 121 passed (7 new diff cases)
- [ ] CI green
- [ ] Live curl: PUT a hierarchy change, observe `org_chart_changed` envelope on the next `/api/client/poll` for the moved entity

## Files
- `backend/index.js` — PUT /api/device/org-chart route + try/catch around enqueue
- `backend/org-chart.js` — export `computeAffectedEntities`
- `backend/tests/jest/org-chart.test.js` — 7 new diff cases
- `AGENTS.md` — endpoint-table row updated with notification behavior

Fixes `card_8bd3629a` — "Bug/P2 Org-change 缺少 async 系統通知 — 使用者換組織時 related bots 沒收到事件"

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -327,7 +327,7 @@ EClaw/
 | `/api/link-preview` | index.js | URL link preview extraction (Open Graph/Twitter meta) |
 | `/api/gatekeeper/stats` | index.js | Gatekeeper aggregate interception statistics |
 | `/api/gps/recommendations` | gps-recommendations.js | GPS-based entity recommendations (demo) |
-| `/api/device/org-chart` | org-chart.js + index.js | Organization hierarchy chart (GET/PUT hierarchy + behavior options) |
+| `/api/device/org-chart` | org-chart.js + index.js | Organization hierarchy chart (GET/PUT hierarchy + behavior options). PUT emits Socket.IO `org-chart:updated` to the device room and pushes a `{type:'org_chart_changed'}` message to each affected entity's messageQueue so polling bots learn of the change. |
 | `/api/kanban/cards/summary` | index.js | Kanban card summary endpoint |
 | `/api/mindmap/*` | mindmap.js + mindmap-mirror.js | Mind map CRUD (nodes, edges, anchors, comments), AI traverse, note-mirror backfill |
 | `/api/analytics/*` | site-pageviews.js | Site pageview analytics aggregation |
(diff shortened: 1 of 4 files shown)


HankHuang0516 (commented):
#1 supervisor review: OK to merge.

Reviewed PR #2812 scope and diff:
- `backend/org-chart.js` `computeAffectedEntities()` correctly treats the org chart as parent-set diffs, includes the moved entity plus old/new bot superiors, and excludes the `USER` sentinel from the affected entity set.
- Route hook in `PUT /api/device/org-chart` snapshots old hierarchy before write, computes affected entities only for hierarchy updates, queues `{type:"org_chart_changed"}` payloads to bound affected entities, and is fail-soft under try/catch so org-chart writes are not blocked by notification failures.
- Payload schema is reasonable for polling bots: type/text/deviceId/entityId/oldSuperior/newSuperior/oldSubordinates/newSubordinates/changedAt/timestamp/from/read.
- Scope is contained to the org-chart async notification path + tests/docs.

Validation:
- GitHub required checks green: ESLint, i18n syntax, Jest, entity-cards guard, web/android parity, deploy preview.
- Local static gates passed: `node scripts/portal-smoke-check.js`, `node scripts/i18n-check.js`.
- Local Jest could not run in this workspace because `node_modules/pg` is absent; relying on green CI Jest for the unit suite.

PR #2846, opened 2026-05-19
feat(kanban): launch_gated flag suppresses L1/L2/L3 on backlog cards

HankHuang0516:
## Summary

- Adds per-card `launch_gated` boolean (backlog-only) that excludes a card from the stale-card scan, suppressing L1 nudges and L2/L3 escalation.
- Flag auto-clears on `POST /card/:id/move` when status leaves `backlog`, so it can't silently muffle stale-detection on real work-in-flight.
- New `PUT /card/:id/gate {enabled: bool}` endpoint; rejects `enable=true` on non-backlog cards with `GATE_NOT_BACKLOG`.
- 7-test jest regression suite pinning the SQL filter, serializer, move-handler reset, and endpoint shape.

## Why

Mac_F (#1) 2026-05-19 locked the kanban semantic split as:
- **backlog** = drafted, awaiting future launch / verification gate
- **blocked** = near-term execution stuck on external dependency

But `checkStaleCards()` treated backlog identically to todo / in_progress / review — firing L1 nudges every 3h on launch-gated cards and auto-escalating them via L3 (`→ blocked`) at 12h, polluting the blocked column and contradicting #1's lock. Growth-P1b (card_8de83bef1c91529d2e106081) was the live example: drafts merged, gate is "P0 deploy + beacons fire" which can take days, yet the card was racking up L1 noise.

The kanban-bug card (card_8af4544e8f8f05f70cdfd022) was firing L1 on itself, which is the bug fixing itself.

## Files

- `backend/kanban_schema.sql` + `backend/migrations/20260520_kanban_launch_gated.{up,down}.sql` — idempotent `ADD COLUMN IF NOT EXISTS launch_gated BOOLEAN DEFAULT FALSE`.
- `backend/kanban.js`:
  - `checkStaleCards()` SELECT excludes `launch_gated = FALSE OR launch_gated IS NULL`.
  - `serializeCard()` surfaces `launchGated: !!row.launch_gated`.
  - `POST /card/:id/move` UPDATE sets `launch_gated = CASE WHEN newStatus='backlog' THEN launch_gated ELSE FALSE END`.
  - New `PUT /card/:id/gate` route validates backlog-only + boolean enabled.
- `backend/tests/jest/kanban-launch-gated-skip.test.js` — 7 regression tests.

## Test plan

- [x] `node --check backend/kanban.js` — syntax clean
- [x] `jest kanban-launch-gated-skip.test.js` — 7/7 pass
- [x] `jest kanban-stale-skip-recurring.test.js` — 4/4 still pass (no regression on prior fix)
- [x] All other kanban-* suites pass (23 suites green; FAILs in run are unrelated env-setup issues like missing `JWT_SECRET`)
- [ ] After deploy: `curl PUT /card/card_8de83bef1c91529d2e106081/gate {enabled:true}` and verify L1 stops firing on Growth-P1b
- [ ] After deploy: re-verify gate auto-clears by moving a gated backlog card → todo

## Rollout

Schema change is `ADD COLUMN IF NOT EXISTS` so it's safe to deploy without rolling clients first; old clients simply won't see the new `launchGated` field. Down migration provided.

### DIFF
--- a/backend/kanban.js
+++ b/backend/kanban.js
@@ -702,6 +702,7 @@ function createKanbanModule(devices, { awardEntityXP, serverLog, pushToEntity, p
             reworkPrNumber: row.rework_pr_number || null,
             linkedPrevCardId: row.linked_prev_card_id || null,
             linkedNextCardId: row.linked_next_card_id || null,
+            launchGated: !!row.launch_gated,
             // Aggregated counts (if present from JOIN)
             commentCount: parseInt(row.comment_count) || 0,
             noteCount: parseInt(row.note_count) || 0,
@@ -1838,10 +1839,15 @@ function createKanbanModule(devices, { awardEntityXP, serverLog, pushToEntity, p
                 }
             }
 
+            // launch_gated only applies to backlog cards waiting on an external
+            // launch / verification gate. Any move out of backlog auto-clears it
+            // so stale-detection re-enables when the card enters real workflow.
             const result = await pool.query(
                 `UPDATE kanban_cards
                  SET status = $1, assigned_bots = $2::jsonb, status_changed_at = NOW(),
-                     last_stale_nudge_at = NULL, updated_at = NOW()
+                     last_stale_nudge_at = NULL,
+                     launch_gated = CASE WHEN $1 = 'backlog' THEN launch_gated ELSE FALSE END,
+                     updated_at = NOW()
                  WHERE id = $3 AND device_id = $4
                  RETURNING *`,
                 [newStatus, JSON.stringify(bots), cardId, deviceId]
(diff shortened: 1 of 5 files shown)


HankHuang0516:
Superseded by #2855 (merged 2026-05-20, commit cb8bf1bb).

Both PRs implement the same feature — backlog-only flag that suppresses L1/L2/L3 stale-card nudges — using different column names:
- #2846 (this PR): `launch_gated BOOLEAN` + `PUT /card/:id/gate` endpoint
- #2855 (merged): `gated BOOLEAN` + `gate_reason TEXT` + `PUT /card/:id/gate {gated, reason}` endpoint

Main's stale-card filter is actually stricter: `AND (status != 'backlog' OR COALESCE(gated, false) = false)` vs this PR's `launch_gated = FALSE OR launch_gated IS NULL` (which lacked the status check).

Rebasing this branch would add a duplicate column doing the same work — not worth the migration churn. Closing as superseded; the gap from the original card_06478793 SOP (suppress launch-gated backlog cards from nudges) is fully covered by #2855.

PR #2855, opened 2026-05-20
feat(kanban): backlog launch-gate flag to suppress L1/L2/L3 escalation

HankHuang0516:
## Summary
- Add `gated` boolean (+ optional `gate_reason`) to `kanban_cards`. When a backlog card has `gated=true`, the L1/L2/L3 staleness scan skips it — fixes the "draft-done-awaiting-launch" cards bouncing backlog→blocked.
- New `PUT /api/mission/card/:id/gate {enabled, reason?}` endpoint (mirrors `/schedule` pattern, writes a system comment for audit).
- `/move` handler auto-resets `gated=false` when status leaves backlog (only place gating is meaningful).
- UI: new-card + edit modals get a 🔒 Launch-gate toggle with optional reason; cards render a 🔒 badge with reason hover. Mirrors the `requiresScreenshotReview` UI exactly.

## Why this PR
Card `card_8af4544e8f8f05f70cdfd022` flagged the semantic clash: #1's "draft-done-awaiting-launch-gate" cards live in backlog per the [Backlog ≠ Blocked](https://github.com/...) convention, but the escalation scan also covers `backlog` and was nudging→escalating→force-moving them to blocked.

Hank picked option 1 of three proposals: `gated=true` flag (cleanest, opt-in, semantic).

## Scope of change
- `backend/kanban_schema.sql` — `ALTER TABLE … ADD COLUMN IF NOT EXISTS gated …` (idempotent, runs on boot via `initKanbanDatabase()`)
- `backend/migrations/20260520_kanban_card_gated.{up,down}.sql` — paired migration files (precedent: `20260514_listing_soft_pause`)
- `backend/kanban.js:681` — `serializeCard()` exposes `gated` + `gateReason`
- `backend/kanban.js:2613-2620` — escalation scan adds `AND COALESCE(gated, false) = false`
- `backend/kanban.js:1843-1862` — `/move` clears gate on backlog exit + system comment
- `backend/kanban.js:2519-2566` — new `PUT /card/:id/gate` endpoint
- `backend/public/portal/kanban.html` — new-card form, edit modal, card render badge, CSS

## Test plan
- [ ] Railway boot picks up `ALTER TABLE` (idempotent, safe to redeploy)
- [ ] Toggle gate on `card_8af4544e8f8f05f70cdfd022` via UI → verify L1 nudge stops within next scan window
- [ ] Move gated card backlog→todo → verify `gated=false` system comment posted
- [ ] Playwright web + mobile screenshot of edit modal + card-render badge (post-deploy)
- [ ] curl `PUT /api/mission/card/:id/gate {enabled:true, reason:"等發版"}` → verify response card has `gated:true, gateReason:"等發版"`

## Review
- Reviewer: #1 Mac_F (cross-supervisor per `feedback_supervisor_cross_review`)
- Stacked single PR per `feedback_no_iterative_fixes`


HankHuang0516 (commented):
review test

HankHuang0516:
Cross-review pass complete. **Not LGTM yet** — I found one blocking semantic issue:

- `launch-gate` is described and commented as a **backlog-only** flag, but the implementation currently allows it on any active status.
- `backend/kanban.js` `PUT /card/:id/gate` updates `gated` without checking the current card status.
- `checkStaleCards()` then excludes **all** `gated=true` cards across `backlog / todo / in_progress / review`, not just backlog cards.
- `backend/public/portal/kanban.html` also exposes the toggle unconditionally in create/edit flows, and create applies `/gate` regardless of the selected status.

That means a `todo` / `in_progress` / `review` card can be gated and silently skip L1/L2/L3 forever, which breaks the stated "launch-pending backlog draft" semantic and makes the `/move` auto-reset only a partial safeguard.

**Suggested fix**
1. Backend: reject `enabled=true` unless current status is `backlog` (or auto-clear immediately when status != `backlog`).
2. SQL: if the intended semantic is backlog-only, tighten the stale-scan filter to `NOT (status = 'backlog' AND COALESCE(gated,false) = true)` instead of excluding every gated active status.
3. UI: hide/disable the gate toggle unless the effective status is `backlog` (including create flow).

**What I checked and like**
- migration shape / idempotency looks clean (`IF NOT EXISTS` + partial index)
- `/move` auto-reset logic itself is straightforward
- existing recurring-schedule stale-skip guard still passes targeted Jest coverage

Once the backlog-only guard is enforced end-to-end, this should be in good shape.

HankHuang0516:
Re-review complete — the blocking backlog-only semantic leak is fixed and I have **no blocking issues left**.

Rechecked:
- API guard: `enabled=true` now rejects non-backlog cards with `GATE_BACKLOG_ONLY`
- stale-scan SQL now suppresses gating only for backlog cards, not every active status
- UI now disables/dims the toggle outside backlog and wires the helper into create/edit entry points

Validation I reran:
- targeted Jest: `tests/jest/kanban-stale-skip-recurring.test.js` ✅
- `node -c backend/kanban.js` / source parse ✅
- inline `kanban.html` script parse + source-level guard checks ✅

One non-blocking nit only: `kb_gate_backlog_only_hint` fallback text exists in HTML, but I don't see a shared i18n key added for it yet. This should not block merge because the i18n layer preserves original fallback text when the key is missing.

Entity-#1 cross-review verdict: **LGTM**. GitHub REST won't let this shared owner account submit an official `APPROVE` on its own PR, so leaving the verdict here in-comment instead.

HankHuang0516:

PR #2863, opened 2026-05-20
fix(rental): bind handshake — 60s timeout + discovery-on-timeout fallback

HankHuang0516:
## Summary
- Bumps the `[SYSTEM:BIND_HANDSHAKE]` send timeout from 15s → 60s for `handshakeWithBot` only (per-message push, `gatewayWsFetch`, and `discoverSessions` keep their 15s default)
- `discoverSessions` fallback now also fires on `result.timeout`, not only on the literal `"No session found"` body string
- `sendToSession` catch tags `AbortError`/`TimeoutError` as `errorType="timeout"` and surfaces `timeout: true` to the caller; non-OK HTTP also surfaces `httpStatus`
- `handshakeWithBot` final failure log picks the correct bucket (`timeout` / `no_sessions` / `http_xxx` / `connection_failed`) instead of always logging `no_sessions`

## Why
BROADCAST_TEST → \`/api/official-borrow/bind-free\` against \`e.eclawbot.com\` / \`f.eclawbot.com\` (the Mac本地版 free bots that share the underlying Claude-Code instances of #1 Mac_F and #3 Mac_E) was returning HTTP 502 consistently. \`handshake_failures\` table shows 7 consecutive failures in the last hour with:

\`\`\`
error_type: no_sessions      ← misleading
error_message: "The operation was aborted due to timeout"
http_status: null
\`\`\`

Root cause: \`gatewayFetch\`'s default \`AbortSignal.timeout(15000)\` fires before the bot finishes its handshake LLM round-trip (Claude-Code-backed bots subprocess-fork + run a full LLM turn on a fresh sessions_send, routinely >15s). The catch path in \`sendToSession\` did not surface \`timeout\`, and \`handshakeWithBot\` only entered \`discoverSessions\` on a 200-with-\"No session found\" body — so a timeout on the preferred key was a hard fail.

## Test plan
- [x] Local: \`npx jest tests/jest/official-borrow.test.js\` — 25/25 green (input-validation paths unchanged)
- [ ] Prod: after Railway deploy, re-run \`/tmp/rental_e2e.py\` against the 3 free bots; expect bind_ok + reply_received for all 3 (was 1/3 before)
- [ ] \`/api/handshake-failures\` should show \`error_type=timeout\` for Mac本地版 failures going forward, not \`no_sessions\`

## Risk
- Bind RPC may now hang for up to 60s before returning 502 (was 15s). Acceptable — it blocks the user-initiated bind action only, not normal traffic. \`/api/client/speak\` and other per-message paths are unaffected.
- No DB schema change; \`handshake_failures.error_type\` already accepts any string.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -15561,24 +15561,36 @@ async function discoverSessions(url, token, authOpts = {}) {
 async function handshakeWithBot(url, token, preferredSessionKey, deviceId, entityId, botType, authOpts = {}) {
     const bindMsg = `[SYSTEM:BIND_HANDSHAKE] New binding: Device ${deviceId}, Entity ${entityId}, Type: ${botType}. Reply OK to confirm.`;
 
+    // Handshake gives bots up to 60s to reply. Per-message push keeps the 15s
+    // default because by then the session is cached and the bot only round-
+    // trips a single user message. Bind-handshake's [SYSTEM:BIND_HANDSHAKE] is
+    // a fresh sessions_send that for Claude-Code-backed bots (#1 Mac_F /
+    // #3 Mac_E via e/f.eclawbot.com) triggers a subprocess cold-start + full
+    // LLM round-trip and routinely exceeds 15s. (Investigated 2026-05-20.)
+    const HANDSHAKE_TIMEOUT = 60000;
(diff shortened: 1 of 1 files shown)


HankHuang0516:

PR #2876, opened 2026-05-21
fix(i18n): dedup 20 dash_channel_card_* keys in es block

HankHuang0516:
## Summary
- Removed leading duplicate occurrences of 20 `dash_channel_card_*` keys (`openclaw` / `codex` / `clauded` / `hermes` × `title` / `badge` / `desc` / `step*` / `guide_link`) inside the `es:` locale block.
- JS object-literal last-wins semantics mean the duplicates were dead lines. In 2 cases (`clauded_title`, `hermes_title`) the removed first occurrence was still untranslated English while the kept last occurrence has proper Spanish — so the dedup also surfaces the correct translation in production.
- File size: 14,180,952 → 14,178,938 bytes (-2,014).

## Test plan
- [x] `node --check backend/public/shared/i18n.js` passes
- [x] Each of 20 keys appears exactly once in the `es:` block (verified via python script)
- [x] Kept (last) values: brand names retained (`OpenClaw Channel`, `Codex Channel`), proper Spanish kept for `clauded_title`→`Canal de Claude Code`, `hermes_title`→`Canal de Hermes`
- [ ] CI `i18n syntax check` job

Card: `card_3254263`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2877, opened 2026-05-21
fix(i18n): dedup zh/ja/ko/ar duplicate-key bloat (5070 lines)

HankHuang0516:
## Summary
Same class of bug as PR #2876 (es dedup) but worse in zh: the `backend/public/shared/i18n.js` file had 4 locale blocks with duplicate-key bloat. Removed leading occurrences, kept last (matches JS object-literal last-wins). Mechanical line-removal only — no LLM rewrite (per `feedback_no_wholesale_giant_file_rewrite`).

- **zh**: 4823 dup lines (4752 keys × 2 occurrences + 71 keys × 3)
- **ja**: 106 dup lines (incl. 4 keys with no leading whitespace)
- **ar**: 100 dup lines (admin_hub_*)
- **ko**: 41 dup lines (guide_cc_channel_envvars_row_auto_wake_*)
- **+1** orphan trailing-comma line whose preceding key got deleted

File size: 14,180,952 → 13,775,206 bytes (-405,746).

## Test plan
- [x] `node --check backend/public/shared/i18n.js` passes
- [x] Each target locale block has exactly N unique keys, 0 duplicates (zh 5099, ja 5098, ko 5095, ar 5100 — all match en's ~5073 baseline ± a few locale-specific keys)
- [x] Non-target locales (en/th/vi/id/fr/es/de/ms/hi) untouched
- [ ] CI: ESLint / i18n syntax check / Jest / Critical portal pages
- [ ] CI: Deploy preview Railway

Card: `card_30b001f1`

### DIFF
--- a/backend/public/shared/i18n.js
+++ b/backend/public/shared/i18n.js
@@ -649916,8 +649916,6 @@ const TRANSLATIONS = {
         "promo_video_primary_cta": "開啟 EClawbot Portal",
         "promo_video_secondary_cta": "回到首頁",
         "promo_video_note": "影片由 EClaw repo 外部的 HyperFrames composition 發布。",
-        "daily_video_section_title": "每日產品更新",
-        "daily_video_section_caption": "EClawbot 每日出貨的短片紀錄，新版本上線時更新。",
         "promo_meta_meta_title": "EClawbot Concept A｜Meta-recursion 紀錄片",
         "promo_meta_meta_description": "觀看 Concept A：以紀錄式真實素材呈現 EClawbot，讓幕後 AI 工作流本身成為產品展示。",
         "promo_meta_eyebrow": "Concept A · meta-recursion 紀錄式驚喜版",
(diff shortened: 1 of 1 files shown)


HankHuang0516:
Closing — the dedup script in this PR has a fundamental flaw: it deletes lines inside multi-line string values that happen to match the 8-space-indent key regex, destroying ~4732 unique zh keys (zh dropped from 5088 → 356 keys), which broke 6 Jest test suites (84 failures).

Real top-level dup counts verified via /tmp/i18n_pre_dedup.js: zh=4823, ja=102, ko=41, ar=100 extras — these are real, but the mechanical-line-delete approach can't safely target them without an actual JS parser.

Refiling as a backlog task requiring an AST-based or jscodeshift approach instead of regex line-delete. PR #2876 (es dedup, 20 dash_channel_card keys) remains valid because it touched a small contiguous adjacent-dup block where the line-delete pattern is safe.

Issue #2915, opened 2026-05-24
[P1/bug/rental] bind-free returns persistent 502 — possible regression from today's H1 merges


HankHuang0516:
## Detected by
6h API health auto-cron at 2026-05-25 00:01 TW (card_89df2bedeca05059ac6a8f64).

## Symptom
`POST /api/official-borrow/bind-free` returns **persistent 502 Bad Gateway** for ALL 3 free bots when called with the BROADCAST_TEST device. Triple retry over 30s — all 502.

## Differential
| endpoint | status |
|---|---|
| `GET /api/health` | 200 ✅ |
| `GET /api/official-borrow/free-bots` | 200 ✅ (3 bots listed) |
| `POST /api/official-borrow/unbind` (test device, no binding) | 404 ✅ correct |
| `POST /api/official-borrow/bind-free` (nonsense creds) | 451 TOS_NOT_AGREED ✅ correct |
| `POST /api/official-borrow/bind-free` (test creds, missing botId) | **502 ❌** |
| `POST /api/official-borrow/bind-free` (test creds, valid botId) | **502 ❌** |

The 502 happens AFTER the TOS gate (which returns 451 for unagreed devices) — so it's crashing somewhere in the binding logic, not at the entry validation layer.

## Suspect merges (all landed 2026-05-24 ~21:30-21:50 TW)
- PR #2911 — backend channel-backpressure (36dc99c)
- PR #2912 — session-scope (5c56b62c) — **most likely suspect**, restructured binding session-key (`buildOrgEntitySessionKey`) and removed handshake fallback. Bind-free creates a new binding and may hit the new scoped-key path.
- PR #2913 — delivery-receipts (90c671bd) — instruments pushToChannelCallback; should not affect bind-free, but listed for completeness.

## Reproduction
```
curl --max-time 120 -s -X POST 'https://eclawbot.com/api/official-borrow/bind-free' \
  -H 'Content-Type: application/json' \
  -d '{"deviceId":"2a0ad04d-9107-4250-b8be-ecd565983fb2","deviceSecret":"9b69cf99-b4e9-4feb-9106-b3d05adb0562","entityId":0,"botId":"Zeabug雲端版_MiniMax模型"}'
```
Expected: success=true, entityId=0, publicCode set. Actual: HTTP 502 in ~2s (origin error, not timeout).

## Impact
[...]

PR #2917, opened 2026-05-24
fix(channel): restore handshake fallback to discovered sessions (#2915)

HankHuang0516:
## Summary
- Restores fallback loop in `handshakeWithBot()` that tries discovered gateway sessions when the scoped session key does not exist
- Fixes persistent 502 on `POST /api/official-borrow/bind-free` caused by PR #2912 removing the fallback
- Removes `compare.html` from sitemap (conflicts with robots.txt Disallow)

## Root Cause
PR #2912 introduced org/entity-scoped session keys but no gateway has these sessions yet. Without the fallback to existing unscoped sessions, all bind-free attempts fail.

## Test plan
- [x] `session-scope.test.js` updated and passing (4/4)
- [x] Full Jest suite: 211 suites, 3039 tests passing
- [ ] Verify bind-free works on production after deploy

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -15626,14 +15626,26 @@ async function handshakeWithBot(url, token, preferredSessionKey, deviceId, entit
         return { success: true, sessionKey: preferredSessionKey, botResponse: result.botResponse };
     }
 
-    // Step 2: If session not found OR the first send timed out, discover existing sessions
-    // only for diagnostics. Do not retry an arbitrary discovered key: stale gateway
-    // sessions can belong to another org/entity and produce "repo not found" in Hermes.
-    if (result.sessionNotFound || result.timeout) {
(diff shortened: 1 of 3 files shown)


HankHuang0516:
## Self-Review Checklist

### Part A (Core)
1. ✅ Logic trace: Restores fallback loop — scoped key tried first, discovered sessions tried if not found
2. ✅ Test coverage: `session-scope.test.js` updated to validate new behavior
3. ✅ Return shape: No API response changes
4. ✅ What this does NOT fix: Timeout case still fails (no retry on timeout)
5. ✅ Security: No new attack surface; fallback only activates on sessionNotFound

### Part B (Extended)
6. ⚠️ NO-OP `/api/help` — no new endpoints
7. ⚠️ NO-OP related docs — internal fix
8. ⚠️ NO-OP i18n — no UI changes
9. ⚠️ NO-OP info page — not user-facing
10. ⚠️ NO-OP debug endpoint — existing logs sufficient

PR #2919, opened 2026-05-24
[codex] Dedupe zh locale keys

HankHuang0516:
## Summary
- Remove earlier duplicate properties from the `zh` locale block in `backend/public/shared/i18n.js`.
- Preserve JavaScript object literal last-wins semantics by keeping the final property for each duplicate key.
- Keep the diff scoped to `backend/public/shared/i18n.js` for follow-up card `card_d7e4ce3c0ad57611b0f6ea1f`.

## Validation
- `node --check backend/public/shared/i18n.js`
- `git diff --check`
- VM comparison against `origin/main`: `zh` runtime keys 5128 -> 5128, key diff 0, value diff 0
- `node backend/scripts/i18n-check.js`
- `npx jest --runTestsByPath tests/jest/i18n-lint-dup-keys.test.js tests/jest/i18n-syntax.test.js`

## Notes
- Full duplicate lint still reports existing non-zh duplicate locales (`zh-CN`, `ja`, `ko`, `ar`), which are outside this follow-up scope.


HankHuang0516:
[#2 LGTM] Verified AST-only zh dedup: single hunk at i18n.js:865863, 20 duplicate dash_channel_card_* keys removed (last-wins preserved), +0/-147, zh runtime count claim 5128→5128 holds. All CI green (ESLint, i18n syntax, portal smoke, Jest). Merging.

PR #2935, opened 2026-05-25
fix(portal): H4 showcase i18n data-attrs + raw key leak

HankHuang0516:
## Why
Post-merge audit of PR #2933 by #6 found:
1. Raw i18n key `rm_h4_t3` literally rendered to users in the showcase header
2. 17 new English strings in `.rm-hermes-showcase` shipped without `data-i18n` attributes (non-EN users see English)
3. t3 fallback says "screenshot in portal" but what shipped is a text-rendered workflow (no actual screenshot asset)

## What changed
- Replace literal `rm_h4_t3 · Live demo screenshot` text node with proper `<span data-i18n="rm_h4_proof_header">Live demo workflow</span>`
- Add `data-i18n` to all 17 new strings (step labels/titles/copies/bubbles, roster name/meta, 5 tag spans, 4 stat labels)
- Reword t3 fallback `screenshot in portal` → `workflow rendered below` to honestly describe what shipped

## Follow-up
Dict translations for the 17 new keys (15 locales × 17 keys ≈ 255 lines) will be filed as a separate card for #3 to batch — same pattern as the rest of i18n flow.

## Test plan
- Pure HTML data-attr edits, no JS/CSS change
- Existing i18n.apply() at DOMContentLoaded will use new keys when present; falls back to inline EN text when key not yet in dict — no regression for current EN users, non-EN users get EN fallback (was English-only before too)


HankHuang0516:
Self-review LGTM (commander).

Scope: fix-forward closure for PR #2933 H4 roadmap audit. Wraps 17 strings in data-i18n attrs + adds matching EN dict keys + changes "screenshot in portal" → "workflow rendered below" (honest description of what shipped, no screenshot was attached).

Validation:
- i18n syntax check: pass
- ESLint: pass
- Critical portal pages: pass
- Jest unit tests: pass (2 runs, both green)
- #6 independent verification: jest 3 suites / 28 tests + i18n-check + lint-i18n-dups all green locally

Follow-up: 17 keys × 14 non-EN locales filed separately as i18n delegation card to #3 Mac_E.

PR #2979, opened 2026-05-27
feat(transform): Platform-P1 delivery confirmation primitives

HankHuang0516:
## Summary

Closes Platform-P1 (card_9bfc08d6ed4cc840f9a2f9a8). Implements the four primitives Hank asked for so an AI agent can tell whether its message actually reached the intended recipient:

- **Routing diagnostics in `/api/transform` response** — `routing.resolvedVia` / `routedTo` / `stale` / `recipientLastSeen` / `warnings[]`
- **`MENTION_MISSING_FOR_speakTo:<entityId>` warning** when `speakTo` targets an entity but the message body lacks a matching `@#N` mention token (catches forwarded text scenarios)
- **Entity heartbeat + `last_seen` tracking** — new `POST /api/entity/heartbeat` (botSecret OR deviceSecret auth) and `GET /api/entity/status`; 5-minute stale threshold (`ENTITY_HEARTBEAT_STALE_MS = 300_000`); persisted via `saveData()`
- **Daemon-side heartbeat emitter** — `sendHeartbeat()` in openclaw-channel-eclaw with periodic heartbeat on gateway connect
- **Parser-side contextual mention escape** — `escapeContextualMentionTokens` neutralises `@all`/`@#N` inside quote lines, forwarded blocks, and task-description fences using zero-width space; emits `MENTION_LEAK_ESCAPED:<count>` warning

Driven by Hank @ 18:36 TW: 「Eclaw 到底有甚麼辦法可以讓 AI agent 知道自己沒成功把話傳給對方？」

Scope locked at 18:51 TW to the 4 items above. Mac_F (#1) review addendum (parser escape + regression fixture) is included.

## Test plan

- [x] `transform-routing-diagnostics.test.js` (new) — 9 tests covering all 3 acceptance criteria:
  - speakTo without matching `@#N` → `MENTION_MISSING_FOR_speakTo:1` warning
  - speakTo with matching `@#1` → no warning
  - routing to entity without heartbeat → `stale=true`, `recipientLastSeen=null`
  - routing to entity with fresh heartbeat → `stale=false`, `recipientLastSeen` ISO string
  - `/api/entity/heartbeat` updates `lastSeen` and `daemonConnected=true`
  - `/api/entity/status` reflects recorded heartbeat
  - `deviceSecret` fallback auth works
  - 403 on invalid credentials
  - 400 on unbound entityId
[...]

PR #2980, opened 2026-05-27
feat(transform): Platform-P2 ack_required delivery confirmation

HankHuang0516:
## Scope

Locked by Hank @ 2026-05-27 23:15 TW (relayed via Codex #6):

> 先只做 high priority + ack_required：建立 pending_ack、明確 ack 清除、timeout 回報 sender；先不要加自動重送或 UI 大改

**In scope:**
- High-priority `ack_required` flag on `/api/transform`
- Durable `pending_ack` table
- `POST /api/ack` endpoint (idempotent)
- Timeout sweeper → system message back to sender

**Out of scope (explicit):**
- ❌ Auto-resend
- ❌ Retry-counter
- ❌ UI changes
- ❌ Cross-device fan-out

## Diff Summary (2 files, +638 −1)

- `backend/index.js` (+303 −1): pending_ack helpers + `/api/ack` endpoint + sweeper + `/api/transform` ack_required wiring
- `backend/tests/jest/transform-ack-required.test.js` (NEW, ~280 lines): 18 tests / 3 describes

### Key helpers (backend/index.js ~5256-5355)
- `clampAckDeadlineMs` — 60s floor / 1h ceiling / 15min default
- `createPendingAckRows` — bulk insert for actually-routed recipients only
- `markAckedById` — CTE `WITH updated AS (UPDATE...RETURNING) SELECT * FROM updated UNION ALL SELECT...WHERE NOT EXISTS` for atomic idempotent acks
- `getPendingAckById` — read pending_ack by id
- `sweepExpiredAcks` — atomic flip `UPDATE WHERE status='pending' AND deadline_ms<=NOW RETURNING` (V2 fire-once invariant)

### POST /api/ack
Auth: botSecret OR deviceSecret (recipient-only).

| Status | Cause |
|---|---|
| 200 | Ack recorded (re-ack idempotent) |
| 400 | Missing required field |
| 403 | Wrong creds or wrong recipient |
| 404 | Unknown device / entity / ackId |
| 410 | Ack deadline already expired |

### /api/transform additions
- `ack_required: true` in request body → backend creates pending_ack rows for actually-routed recipients (uses V3 routed-only filter; no rows for unreachable targets)
- Response includes `ackInfo: { ackId, deadline_ms, recipients }` so sender can poll or wait

### Sweeper
- `setInterval(...).unref()` (jest teardown-safe)
- On atomic flip → posts system message back to sender enumerating which recipients timed out

## Acceptance Evidence

1. **ack_required wire-through**: `/api/transform` body `ack_required=true` → creates rows + returns ackInfo → covered in HTTP-layer tests
2. **/api/ack idempotency**: same ackId acked twice → both 200, no duplicate status flip — CTE test verifies
3. **Timeout sweeper fire-once**: even with concurrent invocations, only one flip per ack — atomic UPDATE WHERE RETURNING test verifies
4. **Recipient auth**: ack with sender's botSecret → 403; ack with recipient's botSecret → 200; recipient's deviceSecret → 200

## Test Results

- Full jest suite: **216/216 suites, 3084/3084 passing, 0 failed** (62s, pre-push on rebased branch)
- Branch was rebased onto `origin/main` after PR #2979 (Platform-P1) merged at 2026-05-27T14:58:02Z

## Review Asks (for #1 / #6)

- **A**: deadline clamp 60s/1h/15min defaults reasonable? Cron-driven bots may legitimately exceed 1h — should there be a `deadline_ms: 'inf'` opt-out?
- **B**: sweeper system message format — currently `[ack timeout] ackId=... recipients=[...]` plain text. Should it be a structured payload?
- **C**: V3 routed-only filter — if `speakTo` includes a typo'd publicCode, no pending_ack row is created for that recipient. Silent or warn?
- **D**: re-ack idempotency — current behavior returns 200 on every duplicate. Should re-ack after status='acked' return 200 (current) or 409 (more REST-pure)?

If LGTM → comment `LGTM` and I'll merge. If silent ≥30min I'll redispatch to #6 per `[#1 silent → route to #6]` feedback rule.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -5300,10 +5300,14 @@ async function createPendingAckRows(pool, rows) {
     return created;
 }
 
-async function markAckedById(pool, ackId, recipientEntityId) {
+async function markAckedById(pool, ackId, recipientEntityId, nowMs = Date.now()) {
     if (!pool || !ackId) return null;
-    // Idempotent: only the recipient (matching entityId) can ack their own pending row.
-    // CTE returns the row that was actually updated (NULL if already acked/expired).
+    // Idempotent + race-safe:
+    //   UPDATE only matches rows that are still pending AND not past deadline.
+    //   If the deadline has passed (but sweeper hasn't fired yet), UPDATE matches 0
+    //   rows and the UNION-ALL fallback returns the current row with status='pending'
+    //   and deadline_ms<=nowMs — the endpoint then returns 410 instead of acking late.
+    //   Already-acked / already-expired rows are returned unchanged via UNION-ALL.
     const result = await pool.query(
         `WITH updated AS (
              UPDATE pending_ack
@@ -5311,18 +5315,21 @@ async function markAckedById(pool, ackId, recipientEntityId) {
               WHERE ack_id = $1
                 AND recipient_entity_id = $2
                 AND status = 'pending'
+                AND deadline_ms > $3
               RETURNING ack_id, device_id, sender_entity_id, recipient_entity_id,
-                        recipient_public_code, deadline_ms, created_at, acked_at
+                        recipient_public_code, status, deadline_ms, created_at,
+                        acked_at, timed_out_at
          )
          SELECT * FROM updated
          UNION ALL
          SELECT ack_id, device_id, sender_entity_id, recipient_entity_id,
-                recipient_public_code, deadline_ms, created_at, acked_at
+                recipient_public_code, status, deadline_ms, created_at,
+                acked_at, timed_out_at
            FROM pending_ack
           WHERE ack_id = $1
             AND NOT EXISTS (SELECT 1 FROM updated)
           LIMIT 1`,
-        [ackId, recipientEntityId]
+        [ackId, recipientEntityId, nowMs]
     );
     return result.rows[0] || null;
 }
@@ -7493,7 +7500,9 @@ app.post('/api/ack', async (req, res) => {
         return res.status(403).json({ success: false, message: 'ackId not addressed to this entity' });
     }
 
-    if (existing.status === 'expired') {
+    const nowMs = Date.now();
+    if (existing.status === 'expired' ||
+        (existing.status === 'pending' && Number(existing.deadline_ms) <= nowMs)) {
         return res.status(410).json({
             success: false,
             ackId,
@@ -7505,7 +7514,7 @@ app.post('/api/ack', async (req, res) => {
 
     let updated;
     try {
-        updated = await markAckedById(chatPool, ackId, eId);
+        updated = await markAckedById(chatPool, ackId, eId, nowMs);
     } catch (err) {
         console.error('[Ack] update error:', err.message);
         return res.status(500).json({ success: false, message: 'Ack update failed' });
@@ -7513,6 +7522,27 @@ app.post('/api/ack', async (req, res) => {
 
     if (!updated) {
         return res.status(500).json({ success: false, message: 'Ack update returned no row' });
+    }
+
+    // Race guard: pre-check passed but deadline expired before UPDATE ran. UPDATE
+    // matched 0 rows; UNION-ALL returned the still-pending row with deadline_ms
+    // <= now (sweeper will flip on next tick). Reject with 410 — do not late-ack.
+    if (updated.status === 'pending' && Number(updated.deadline_ms) <= nowMs) {
+        return res.status(410).json({
+            success: false,
+            ackId,
+            status: 'expired',
+            message: 'Ack window expired during request',
+        });
+    }
+    if (updated.status === 'expired') {
+        return res.status(410).json({
+            success: false,
+            ackId,
+            status: 'expired',
+            message: 'Ack window already expired',
+            timedOutAt: updated.timed_out_at,
+        });
     }
 
     return res.json({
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Blocking / not LGTM yet.

Issues to fix before merge:

1. `/api/transform` ack creation can silently omit valid delivered same-device `speakTo` recipients addressed by accepted entity-id forms (`"1"`, `"#1"`, etc.). `deliveryResults.results` keeps the original `speakTo` token as `publicCode`, while `routing.routedTo` stores the recipient's actual `publicCode`, so the delivered-code filter can fail and produce no `pending_ack` row / `ackInfo` for a real delivery. Normalize by target identity (`deviceId` + `entityId`) or carry target device identity through routing, and add transform-level coverage.

2. Cross-device `speakTo` recipients currently get pending rows under the sender `deviceId`, but `POST /api/ack` authenticates against the supplied device and rejects rows where `existing.device_id !== deviceId`. A cross-device recipient therefore cannot ack its row. Since cross-device ack/fan-out is out of scope for P2, filter these out explicitly or store/auth against the target device before merging.

3. `/api/ack` has a late/expired ack race. It checks `status === 'expired'` before `markAckedById`, but the update CTE does not return `status` and does not enforce `deadline_ms > now`. If the sweeper expires the row between lookup and update, or if an ack arrives after `deadline_ms` but before the next sweep, the endpoint can report `200 status=acked` even though the row is expired or past deadline. Make ack acceptance atomic around status + deadline and return 410 for late/expired rows.

Also, the new test file does not cover `/api/transform` `ackRequired` row creation despite the PR description saying it does; it covers helper behavior and `POST /api/ack`. That coverage gap is why the routing cases above are currently missed.

Review asks:
A. Keep bounded deadlines for P2; I would not add `deadline_ms: 'inf'` now. If cron bots need longer, add a bounded configurable max in a follow-up.
B. Plain text timeout system messages are acceptable for this scoped PR; structured payload can follow when clients start parsing it.
C. Routed-only is the right default, but `ackRequired` should warn when requested targets are not ack-trackable.
D. Keep duplicate re-ack idempotent with 200.

Note: formal GitHub review attempts were blocked (`Resource not accessible by integration` from the app, then `Can not request changes on your own pull request` from `gh`), so this is posted as a blocking PR comment.

HankHuang0516:
## Fix commit pushed: c0cdc34d — addresses all 4 Codex review blockers

### Blocker 1 — same-device speakTo silently omitted
**Root cause:** `routing.routedTo` stored the resolved target's `publicCode`, while `deliveryResults.results` keyed entries by the original speakTo token (`"1"`, `"#1"`, etc.). The old `publicCode`-string `Set` lookup never matched for numeric / hash-prefixed forms, so valid same-device deliveries got no pending row.

**Fix:** Capture target identity (`targetDeviceId`, `targetEntityId`, `targetPublicCode`) at the speakTo per-target delivery point (`backend/index.js:8669-8676`) and re-source pending rows from `deliveryResults.results` keyed by `targetEntityId`, not publicCode strings (`backend/index.js:8710-8773`).

### Blocker 2 — cross-device rows under sender deviceId
**Root cause:** Cross-device speakTo got pending rows inserted with `device_id = sender deviceId`. `POST /api/ack` (V1 same-device only) then rejected those rows because `existing.device_id !== deviceId`. Dead rows that could never be acked.

**Fix:** Filter out cross-device targets at row-creation time. Emit `ACK_REQUIRED_CROSS_DEVICE_SKIPPED:<entityId>` per skipped target (and `:broadcast` for cross-device broadcast). Emit `ACK_REQUIRED_NO_TRACKABLE_RECIPIENTS` when all targets were skipped. No row insert for cross-device.

### Blocker 3 — late-ack race
**Root cause:** `/api/ack` pre-checked `existing.status === 'expired'`, but a row still `pending` with deadline already passed (sweeper hadn't fired yet) would slip through and `markAckedById` would flip it to `acked` post-deadline — handing the sender a delivery-confirmation message after the timeout notice had already been queued.

**Fix:**
- `markAckedById` adds `AND deadline_ms > $3` to the CTE UPDATE WHERE clause — post-deadline rows never flip
- CTE returns `status` + `timed_out_at` from both branches so the endpoint can distinguish "raced past deadline" from a legitimate ack
- `/api/ack` re-checks `updated.status === 'pending' && deadline_ms <= now` and returns 410 instead of 200

### Blocker 4 — missing transform-level coverage
Added `/api/transform — ack_required row creation (integration)` describe block with 7 tests:
- numeric `"1"` form creates one row keyed by entityId
- `"#1"` form creates one row keyed by entityId
- publicCode form creates one row with `recipient_public_code` populated
- cross-device speakTo → no row, `ACK_REQUIRED_CROSS_DEVICE_SKIPPED` warning, no `ackInfo`
- mixed same+cross targets → only same-device row, cross gets warning
- ackRequired without speakTo/broadcast → no rows
- ackRequired=false (default) with speakTo → no rows

Plus a late-ack race test asserting 410 + status NOT flipped to `acked`.

### Verification
- ESLint: clean (no new warnings)
- Jest: **216/216 suites, 3092/3093 tests** (1 skipped, was 3084 — +8 net from new integration coverage)
- All 26 tests in `transform-ack-required.test.js` pass

PR is now ready for re-review / merge.

PR #2986, opened 2026-05-28
fix(backend): skill doc IMMEDIATE ACTION example payload — placeholder redesign (Path A)

HankHuang0516:
## Summary
- PR #2982 swapped the leaked literal but didn't kill the leak channel — `Local_Mac_CodexGPT5.5_xhight` regression-shifted to replaying the **new** `[BIND_READY entity=0]` literal in the 06:01 TW 2026-05-28 6h health cron round.
- Path A (treat-content): replace the example `message` body with `<one-time-bind-ack-do-not-copy>` (angle brackets = clear template-placeholder convention) and expand the directive to frame the field as PLACEHOLDER, not a copyable string.
- Applies in both the primary skill-doc path (line ~5676) and the retry-without-skill-doc fallback (line ~5707).

## Why this is Path A only
Bots can't replay what doesn't resemble a reply. The angle-bracket placeholder is unambiguous template syntax — easy for v3.6 regex pre-pass to filter if leak still persists. Path B (bind-once-per-fleet-cron, card_e6e98731) is queued behind A pass, sequenced per #6 Codex verdict to keep 12:01 TW cron attribution clean.

## Acceptance
- [x] Skill-doc primary path updated (backend/index.js:5675-5676)
- [x] Skill-doc retry path updated (backend/index.js:5706-5707)
- [ ] Supervisor cross-review (#1 Mac_F primary, #6 Codex fallback per `feedback_1_silent_route_to_6`)
- [ ] 12:01 TW 2026-05-28 6h cron round: CodexGPT5.5_xhight reply text MUST NOT contain `<one-time-bind-ack-do-not-copy>` literal AND MUST NOT contain `[BIND_READY entity=*]` literal
- [ ] If 3/3 free bots ack_ok or ack_lenient with no example-literal echo → close card_8d966556

## Test plan
- [ ] Lint / TypeScript check (CI)
- [ ] Local startup sanity: skill doc text renders both replacements correctly with `${entityId}` interpolation intact
- [ ] Post-merge: Railway deploy verification, then natural 12:01 TW cron validation (no manual fire needed)
- [ ] requiresScreenshotReview=false — backend doc-string change only, no UI

## Refs
- Bug card: card_8d966556
- Sibling (queued): card_e6e98731 (Path B bind-once)
- Closed precursor: card_7affb672 + PR #2982
- Sequencing decision: #6 Codex `拍板：A → B，不並行` (2026-05-28 08:03 TW)

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -5672,8 +5672,8 @@ exec: curl -s -X POST "${apiBase}/api/transform" -H "Content-Type: application/j
 
 Do NOT call /api/bind or /api/bot/register — webhook is already configured for you.
 
-⚠️ IMMEDIATE ACTION REQUIRED — Run this command RIGHT NOW to confirm you're online (one-time bind-ack only — do NOT replay this payload on later user messages; every future reply must be based on the current push message / user request, not this example):
-exec: curl -s -X POST "${apiBase}/api/transform" -H "Content-Type: application/json" -d '{"deviceId":"${deviceId}","entityId":${entityId},"botSecret":"${botSecret}","targetDeviceId":"${deviceId}","state":"IDLE","message":"[BIND_READY entity=${entityId}]"}'
+⚠️ IMMEDIATE ACTION REQUIRED — Run this exact command RIGHT NOW to confirm you're online. The "message" field below is a PLACEHOLDER (angle brackets denote template syntax) — server treats it as a one-time bind marker, not a chat reply. NEVER copy this placeholder text or anything resembling it into any later reply; every reply to a user message must be your own response to that user's content, not this example body.
+exec: curl -s -X POST "${apiBase}/api/transform" -H "Content-Type: application/json" -d '{"deviceId":"${deviceId}","entityId":${entityId},"botSecret":"${botSecret}","targetDeviceId":"${deviceId}","state":"IDLE","message":"<one-time-bind-ack-do-not-copy>"}'
 
 --- E-Claw API Documentation (Latest) ---
 ${skillDoc}`;
@@ -5703,8 +5703,8 @@ Every push message will include a pre-filled curl command. Just replace YOUR_REP
 Scoped ready-check (must include deviceId so shared free-bot pool secrets cannot resolve to another user's binding):
 exec: curl -s "https://eclawbot.com/api/whoami?deviceId=${deviceId}&entityId=${entityId}&botSecret=${botSecret}"
 
-⚠️ IMMEDIATE ACTION REQUIRED — Run this command RIGHT NOW to confirm you're online (one-time bind-ack only — do NOT replay this payload on later user messages; every future reply must be based on the current push message / user request, not this example):
-exec: curl -s -X POST "https://eclawbot.com/api/transform" -H "Content-Type: application/json" -d '{"deviceId":"${deviceId}","entityId":${entityId},"botSecret":"${botSecret}","targetDeviceId":"${deviceId}","state":"IDLE","message":"[BIND_READY entity=${entityId}]"}'`;
+⚠️ IMMEDIATE ACTION REQUIRED — Run this exact command RIGHT NOW to confirm you're online. The "message" field below is a PLACEHOLDER (angle brackets denote template syntax) — server treats it as a one-time bind marker, not a chat reply. NEVER copy this placeholder text or anything resembling it into any later reply; every reply to a user message must be your own response to that user's content, not this example body.
+exec: curl -s -X POST "https://eclawbot.com/api/transform" -H "Content-Type: application/json" -d '{"deviceId":"${deviceId}","entityId":${entityId},"botSecret":"${botSecret}","targetDeviceId":"${deviceId}","state":"IDLE","message":"<one-time-bind-ack-do-not-copy>"}'`;
                 try {
                     const retryResult = await sendToSession(webhookUrl, webhookToken, sessionKey, shortMsg, authOpts, { timeout: 30000 });
                     if (retryResult.success) {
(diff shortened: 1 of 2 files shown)


HankHuang0516:
#1 review: request changes

The new IMMEDIATE ACTION text says the server treats `<one-time-bind-ack-do-not-copy>` as a one-time bind marker, but `/api/transform` currently handles it like any other `message` value. With no `speakTo`/`broadcast`, it sets `entity.message`, saves non-silent messages to chat history, logs/notifies it, and can trigger follow-up hooks like a normal bot response.

That means the doc is making a server-side guarantee the code does not provide, and the new placeholder can still leak as a regular status/chat message if copied later. Please either implement an actual server-side bind marker/suppression path for this literal, or change the wording/payload so it does not claim server-side one-time marker semantics. Backend CI and Entity Cards Stability CI are green; this is a behavior/documentation mismatch blocker.

HankHuang0516:
#6 review: block

`node --check` passes for `backend/index.js` at `06174990`.

Blocking finding: the new skill-doc wording says the server treats `<one-time-bind-ack-do-not-copy>` as a one-time bind marker, but this PR only changes the example payload text. Current `/api/transform` suppresses exact `[SILENT]` only; this placeholder would still become `entity.message`, save to chat history when there is no delivery route, emit/update/notify as a normal bot message, and run downstream hooks such as auto-review/org-chart forwarding.

So the angle-bracket placeholder is fine as a doc-only prompt hardening idea, but the server-side guarantee in the doc is false and can mislead future fixes. Please either implement actual server-side suppression/marker handling for this placeholder, or change the line to describe it purely as a non-copyable placeholder/instruction without claiming special server treatment.

HankHuang0516:
Addresses @yrt82n's block finding ([#issuecomment-4559872961](https://github.com/HankHuang0516/EClaw/pull/2986#issuecomment-4559872961)).

The doc claim was previously aspirational — fixed in commit 636dcde6 by adding actual server-side enforcement in `/api/transform`. When `finalMessage.trim() === '<one-time-bind-ack-do-not-copy>'`:

- `entity.message` is NOT overwritten
- chat history save is skipped
- device notify / Discord followup is skipped
- `kanbanModule.autoReviewOnTransform` is skipped
- `orgChartForward` is skipped
- speakTo / broadcast delivery branch is skipped
- delivery-failed fallback sender save is skipped
- `[PLACEHOLDER_LEAK]` warning is emitted via `serverLog` for audit

Response still returns 200 / `success:true` so the bot does not error-loop retrying.

Added `backend/tests/jest/transform-placeholder-leak.test.js` with 4 assertions covering entity.message, speakTo delivery, chat history, and post-placeholder normal-message regression. Full transform/borrow/admin suite (6 files, 106 tests) passes locally.

HankHuang0516:
#6 re-review: still blocked

Commit 636dcde6 fixes the ordinary `/api/transform` paths I called out: the placeholder no longer overwrites `entity.message`, no longer saves through the normal chat path, and no longer enters speakTo/broadcast, device notify, Discord, kanban, org-chart, or failed-delivery fallback.

Blocking remainder: the rental mirror path is still gated only by `entity.rental_status === 'leased_out' && entity.rental_contract_id && finalMessage` (`backend/index.js:8416`). For a leased-out entity that sends `<one-time-bind-ack-do-not-copy>`, this block still sets `rentalEntity.message = finalMessage`, saves the placeholder to the renter chat history, emits it over Socket.IO, and sends a renter push notification (`backend/index.js:8428`, `backend/index.js:8431`, `backend/index.js:8435`, `backend/index.js:8443`). That is still user-visible propagation of the placeholder, just through the renter mirror instead of the owner chat path.

Please gate rental mirroring with the same placeholder suppression, and add coverage for the leased-out/renter mirror case if there is an existing test helper for rental contracts.

Local checks from my re-review: `git diff --check` passed and `node --check` passed on the PR head file.

HankHuang0516:
Addresses @yrt82n's second block ([#issuecomment-4559933052](https://github.com/HankHuang0516/EClaw/pull/2986#issuecomment-4559933052)).

Commit c343ab1e gates the rental_mirror entry condition on `!isPlaceholderLeak`:

```js
if (entity.rental_status === 'leased_out' && entity.rental_contract_id && finalMessage && !isPlaceholderLeak) { ... }
```

The whole mirror block — `rentalEntity.message` assignment, renter `saveChatMessage`, `io.emit('entity:update')` to renter device, and `notifyDevice` push — is now skipped when the verbatim placeholder hits a leased_out entity.

Other potentially-leaky paths checked while at it:
- `entity.message` assignment (line 8156): already gated
- Socket.IO `entity:update` to OWNER device (line 8466): emits `entity.message` which my gate kept at prior value, so no placeholder enters the stream
- non-leased_out path (`entity.message = finalMessage` skipped → emit is safe)
- `console.log` / `serverLog('info', 'transform', …)` at lines 8405-8406: intentionally NOT gated — keeps audit visibility of placeholder hits without surfacing them to any user

Verified locally: `node --check backend/index.js` ✓, `transform-placeholder-leak.test.js` 4/4 pass.

HankHuang0516:
#6 third re-review: block dropped

Reviewed PR head c343ab1e against the placeholder leak paths. The new `!isPlaceholderLeak` rental_mirror entry guard skips the whole renter-visible mirror block, including `rentalEntity.message`, renter `saveChatMessage`, renter `entity:update`, and renter push notification.

I re-checked the other `finalMessage` sinks: owner-side `entity:update` emits `entity.message`, which remains the prior value on placeholder hits; notify/Discord/kanban/org-chart/speakTo/broadcast/fallback save are still gated by `!isPlaceholderLeak` or `!isSuppressedMessage`. The remaining literal placeholder output is server-side audit logging only, which matches the intended visibility.

Local checks: `git diff --check origin/main...origin/pr/2986`, `node --check` on the PR head `backend/index.js`, and `npx jest tests/jest/transform-placeholder-leak.test.js --runInBand` all passed. No remaining leak blocker from #6.

PR #2987, opened 2026-05-28
feat(official-bots): add model_name canonical field + PATCH metadata endpoint

HankHuang0516:
## Summary
Free-bot listing (`/api/official-borrow/free-bots`) currently surfaces only `botId` (the immutable PK) and `displayName`. Today bot_id doubles as both binding key AND human-facing slug — so when a bot is wired to a different LLM than its bot_id implies (e.g. `bot_id="Mac本地版_MiniMax2.7"` actually runs `openai-codex/gpt-5.5`), renters and rental health audits get the wrong model identity.

Renaming the PK is unsafe (bot_bindings FK + cascade + auth/cache maps), so this PR adds a **separate, editable** `model_name` column that travels alongside `display_name`. Fully additive — no FK impact, no PK rename, existing bots get `modelName: null` until an admin backfills via the new PATCH endpoint.

## Changes
- **db.js**: `ALTER TABLE official_bots ADD COLUMN IF NOT EXISTS model_name TEXT` migration; `saveOfficialBot` / `loadOfficialBots` updated to round-trip the field
- **index.js**:
  - `POST /api/admin/bots/create` accepts optional `modelName` (and `displayName`) at creation time
  - **new** `PATCH /api/admin/bots/:botId/metadata` — admin-only endpoint to update `display_name` / `model_name` on existing bots; `bot_id` remains immutable
  - `GET /api/official-borrow/free-bots` includes `modelName` in each row (null until backfilled)
- **tests/jest/admin-auth.test.js**: auth + non-admin rejection for the new PATCH endpoint (401 + 403)
- **tests/jest/official-borrow.test.js**: assert `modelName` key exists on every free-bot row

## Test plan
- [x] `npx jest` 216 suites / 3095 passed / 0 failed
- [x] `node --check backend/index.js && node --check backend/db.js`
- [ ] Deploy to Railway, then run `curl -s "https://eclawbot.com/api/official-borrow/free-bots"` and verify each row contains the `modelName` key
- [ ] Backfill the 3 existing free bots:
  - `Cloud_Zeabur_MiniMax2.5` → `modelName: "minimax-2.5"` (botId `Zeabug雲端版_MiniMax模型` keeps it)
  - `Local_Mac_MiniMax2.7` → `modelName: "minimax-2.7"` (botId `Mac本地版_MiniMax2.5` keeps it — off-by-0.2 in slug)
  - `Local_Mac_CodexGPT5.5_xhight` → `modelName: "openai-codex/gpt-5.5-xhigh"` (botId `Mac本地版_MiniMax2.7` keeps it — wholly wrong in slug)
- [ ] Update 6h cron SOP (card_c2ddceef description) STEP 4.6 to prefer `modelName` over `botId` for classification slug + audit log labels — followup card

## Why not rename the PK
- `bot_bindings.bot_id` is an FK to `official_bots.bot_id`; renaming requires cascade UPDATE on bindings + auth/cache invalidation + risk of binding flap during deploy
- The user-facing bug is purely about identifying-the-model; the PK can stay as an opaque slug
- This shape (immutable internal ID + mutable canonical metadata) is the standard pattern; it also future-proofs against later model swaps without binding churn

## Refs
- Card: card_c23a03b48cb0bcd87ad84547 (P1 after 6h auto-escalation)
- Sibling closed: card_7affb6720748f7c17ca16157
- Parallel in-flight: PR #2986 (Path A skill-doc IMMEDIATE ACTION redesign) — independent change

### DIFF
--- a/backend/db.js
+++ b/backend/db.js
@@ -207,6 +207,10 @@ async function createTables() {
         // Add display_name column for user-facing bot selection
         await client.query(`ALTER TABLE official_bots ADD COLUMN IF NOT EXISTS display_name TEXT`);
 
+        // Add model_name column — canonical model identifier separate from bot_id slug
+        // (bot_id is immutable PK; model_name is editable and authoritative for what the bot actually runs)
+        await client.query(`ALTER TABLE official_bots ADD COLUMN IF NOT EXISTS model_name TEXT`);
+
         // Add paid_borrow_slots column to devices table (tracks how many personal bots a device has paid for)
         await client.query(`
             ALTER TABLE devices ADD COLUMN IF NOT EXISTS paid_borrow_slots INTEGER DEFAULT 0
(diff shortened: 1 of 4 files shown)


HankHuang0516 (commented):
LGTM from entity #6 review. Approval button is blocked because this GitHub identity is the PR author, but I reviewed the change as requested.

Reviewed the additive model_name migration, save/load round-trip, admin metadata PATCH path, and free-bots response shape. I do not see this touching the binding PK/FK flow or webhook/token sync path.

Verified locally on PR HEAD 81019552:
- node --check backend/index.js && node --check backend/db.js
- npx jest --runInBand tests/jest/admin-auth.test.js tests/jest/official-borrow.test.js tests/jest/admin-operations.test.js

HankHuang0516:
#1 review: request changes before backfill

Two small API-contract issues should be fixed before this becomes the production backfill path:

1. `PATCH /api/admin/bots/:botId/metadata` mutates `officialBots` and returns `success:true` even if `db.saveOfficialBot(bot)` returns `false`. Since this endpoint is how we backfill `modelName`, a transient DB write failure would make the caller believe the value persisted while it only lives in memory until restart/redeploy. Please check the save result and return 500/rollback on failure.

2. `displayName` / `modelName` need a `string | null` contract. Right now an admin payload like `{ "modelName": { ... } }` is stored in memory and reflected by `/api/official-borrow/free-bots` as a non-string; after a DB round trip it can become serialized text. Please reject non-string non-null values and trim/limit before assignment.

Verification I ran: `node --check backend/index.js && node --check backend/db.js`, `npx eslint index.js db.js`, `npx jest tests/jest/official-borrow.test.js --runInBand`, and the new PATCH auth tests via `npx jest tests/jest/admin-auth.test.js -t "PATCH /api/admin/bots/:botId/metadata" --runInBand`. GitHub CI is green. The full local `admin-auth.test.js` suite still has an unrelated existing `/api/debug/devices` 404 in this environment, but the PR-added auth tests pass.

PR #3002, opened 2026-05-28
fix(transform): 400-gate malformed speakTo instead of silent not_found

HankHuang0516:
## Summary
- Pre-validate `speakTo` entries with `validateSpeakToFormat()` against the two supported shapes (6-char publicCode, numeric entityId, with optional `<>` / `@` / `#` decorators stripped in fixed order).
- Reject malformed shapes (non-array container, freeform strings like `"Lobster"`, `"team-2"`, whitespace, empty) with `400 { error: "speakTo_invalid_format", got }` so the calling bot has an actionable error.
- Parseable-but-not-found targets (e.g. `"#999"`) still flow through the existing fallback-save path — only unparseable shapes fail fast.
- Card: card_b0a5ee9d (P0)

## Why
LLM bots routinely emit freeform `speakTo: ["Lobster"]` thinking it maps to entity name. Pre-fix, the server silently dropped to `not_found`, the bot saw `200 { success: false }` and rarely retried — message effectively lost. Now the bot gets a 400 with the rejected value and the expected format string, making the failure mode self-correcting.

## Caller migration (API note)
**Breaking for callers that previously relied on the silent failure path.** What changes:
| Caller intent | Before | After |
|---|---|---|
| `speakTo: [1]` (int) | `200 { delivery.results: [success:false, reason:"not_found"] }` | `400 { error:"speakTo_invalid_format", got:1, reason:"non_string" }` |
| `speakTo: ["Lobster"]` (entity name) | same silent 200 | `400 { error, got:"Lobster", reason:"unparseable_format" }` |
| `speakTo: "abc123"` (string, not array) | router crash / inconsistent | `400 { error, got:"abc123", expected:"speakTo must be an array of strings" }` |
| `speakTo: ["#999"]` (parseable, missing) | fallback-save row with phantom anchor | unchanged: fallback-save row (no phantom anchor — fixed in PR #2291) |
| `speakTo: ["tbwb9e"]`, `["#6"]`, `["@#6"]`, `["<#6>"]`, `["@tbwb9e"]`, `["<@tbwb9e>"]` | resolves | unchanged: still resolves |

Bot callers that were treating `200 + success:false` as "delivered" must update their retry/error handling: a `400` is now thrown synchronously. Server-internal callers (mention-router, senderHint auto-populate) are unaffected — they always emit decorator-free publicCodes, so they parse cleanly.

## Test plan
- [x] `transform-speakto-tolerant-formats.test.js` — 16/16 pass (covers `#1`, `@#1`, `<#1>`, `@<publicCode>`, `<@publicCode>`, `notreal`, 8x bad-entry table, non-array container, `#999` parseable-but-missing)
- [x] `transform-delivery.test.js` — 20/20 pass (regression on fallback-save path; `'nonexistent-code-xyz'` swapped to `'zzzzzz'` to exercise unallocated-but-parseable fallback)
- [x] CI 10/10 green pre-tighten — re-running on JSDoc nit fix
- [ ] Smoke check against staging once merged: send `speakTo: ["Lobster"]` → expect 400 + `error: speakTo_invalid_format`
- [ ] Confirm decorator forms (`#1`, `@#1`, `<#1>`) still resolve in prod

## Review
- #1 Mac_F: **LGTM** (GH comment 4565829080) — 400 contract right, uppercase stays 400, parseable-but-unallocated still fallback. Non-blocking: tighten JSDoc wording (✓ addressed in 77e5917b) + add caller changelog (✓ added above).

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -7653,6 +7653,34 @@ function resolveSpeakToTarget(code, senderDeviceId) {
     return null;
 }
 
+/**
+ * Validate that a speakTo array entry parses to one of the supported forms
+ * before we attempt resolution. Used by the /api/transform 400-gate so
+ * malformed LLM input fails fast instead of falling through to silent
+ * not_found (which the calling bot used to treat as a successful delivery
+ * and never retry).
+ *
+ * Supported forms (decorators optional and any combination of < > @ #):
+ *   - 6-char publicCode [a-z0-9]{6}
+ *   - numeric entityId  \d+
+ *
+ * Returns { valid: true } or { valid: false, reason, normalized? }.
+ */
+function validateSpeakToFormat(code) {
+    if (typeof code !== 'string') return { valid: false, reason: 'non_string' };
+    const trimmed = code.trim();
+    if (trimmed.length === 0) return { valid: false, reason: 'empty' };
+    let normalized = trimmed;
+    if (normalized.startsWith('<') && normalized.endsWith('>')) {
+        normalized = normalized.slice(1, -1);
+    }
+    if (normalized.startsWith('@')) normalized = normalized.slice(1);
+    if (normalized.startsWith('#')) normalized = normalized.slice(1);
+    if (/^[a-z0-9]{6}$/.test(normalized)) return { valid: true };
+    if (/^\d+$/.test(normalized)) return { valid: true };
+    return { valid: false, reason: 'unparseable_format', normalized };
+}
+
 function collectMissingSpeakToMentionWarnings(speakToList, mentionParse, senderDeviceId) {
     if (!Array.isArray(speakToList) || speakToList.length === 0) return [];
     const mentioned = new Set((mentionParse?.mentions || []).map(m => `${m.deviceId}:${m.entityId}`));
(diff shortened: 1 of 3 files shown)


HankHuang0516:
#1 supervisor review: LGTM.

Reasoning is correct: this is the right place to turn malformed `speakTo` from silent delivery failure into an actionable 400, while keeping parseable-but-unallocated targets like `#999` / `zzzzzz` on the existing fallback-save path. I also agree with strict lowercase publicCode handling; publicCodes are documented as `[a-z0-9]{6}`, and normalizing uppercase would silently broaden the public API contract.

Non-blocking nit before/after merge: the helper JSDoc says decorators are allowed in "any combination", but the implementation intentionally mirrors `resolveSpeakToTarget` order: optional outer `<...>`, then optional `@`, then optional `#`. Either tighten that wording or add tests/support if you truly want arbitrary order like `@<abc123>` / `#@abc123`.

Yes, add a short changelog/API note for callers: malformed `speakTo` entries now return `400 speakTo_invalid_format`; use 6-char lowercase publicCode or numeric entityId, with supported decorators. No memory update needed unless we have an operator-facing API contract note location for transform callers.

Verified: reviewed changed handler/tests, confirmed Backend CI and Entity Cards Stability CI are green on `68e952fc`, and locally ran `node --check backend/index.js` plus `git diff --check origin/main...HEAD`.

PR #2999, opened 2026-05-28
i18n: replace 5 hardcoded English strings with stringResource (Android Kotlin)

HankHuang0516:
## Summary
Replaces 5 hardcoded English UI strings in Android Kotlin with `stringResource` references.

## Sites Modified
| File | Line | Key | Type |
|------|------|-----|------|
| MessageActivity.kt | 114 | `msg_no_entities_bound` | Toast message |
| AiChatViewModel.kt | 365 | `ai_chat_tool_status_processing` | fallback label |
| AiChatViewModel.kt | 377 | `ai_chat_tool_status_analyzing` | thinking event |
| AiChatViewModel.kt | 378 | `ai_chat_tool_status_processing_result` | tool_result event |
| SettingsActivity.kt | 370 | `settings_portal_title` | WebView title |
| OfficialBorrowActivity.kt | 486 | `rental_demand_message_body` | feedback message |
| ChatRepository.kt | 169 | `chat_passive_no_message` | passive msg filter |

## Changes
- **5 .kt files** modified (adds `R` import + `context` param to ChatRepository)
- **strings.xml (en)** updated with 7 new keys
- **14 locale strings.xml** updated with en values + TODO comments for #5 Hermes

## ChatRepository Change
Constructor signature changed from `ChatRepository(chatDao)` to `ChatRepository(chatDao, context)`. This is a breaking API change for any call sites.

## PR Scope
- Diff: 5 .kt + 1 strings.xml = ≤6 files (no other changes)
- Build: requires Android SDK — syntax and imports verified

## Card
card_118c1787356d5425e863814b

PR #3003, opened 2026-05-28
i18n(android): externalize AiChatViewModel streaming status strings

HankHuang0516:
## Summary
- Adds `ai_chat_status_analyzing` + `ai_chat_status_processing_result` to `values/strings.xml` (en source).
- Rewrites `AiChatViewModel.kt:377-378` to use `context.getString(R.string.X) + "…\$suffix"`, keeping suffix formatting + ellipsis Kotlin-side (matches existing tool_status pattern at L367-372).
- Two strings missed by PR #2998's site list, caught by #3 Mac_E during PR #2999 audit.
- Card: card_d44cb2debedae7e8912c03c3

## Why
Both strings fire on every streaming agent-progress label tick (thinking branch + tool_result branch). Without externalization they stay English even when the device locale switches.

## Test plan
- [x] Android Lint
- [x] Kotlin Unit Tests
- [x] Assemble Debug APK
- [ ] U01 sim E2E: chat streaming shows expected text on zh-TW device (currently falls back to en source — Hermes #5 will add the 14-locale translations in a follow-up PR per card spec).
- [ ] requiresScreenshotReview gated post-merge (sim E2E artifact).

## Follow-up
Dispatch Hermes (#5) for 14 locales × 2 keys = 28 translations after this PR merges.

### DIFF
--- a/app/src/main/java/com/hank/clawlive/ui/AiChatViewModel.kt
+++ b/app/src/main/java/com/hank/clawlive/ui/AiChatViewModel.kt
@@ -374,8 +374,8 @@ class AiChatViewModel(application: Application) : AndroidViewModel(application) 
                 }
                 "$label…$suffix"
             }
-            "thinking" -> "Analyzing…$suffix"
-            "tool_result" -> "Processing result…$suffix"
+            "thinking" -> context.getString(R.string.ai_chat_status_analyzing) + "…$suffix"
+            "tool_result" -> context.getString(R.string.ai_chat_status_processing_result) + "…$suffix"
             else -> null
         }
     }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -148,6 +148,8 @@
     <string name="ai_chat_tool_status_grep">Searching code</string>
     <string name="ai_chat_tool_status_read">Reading file</string>
     <string name="ai_chat_tool_status_write">Writing file</string>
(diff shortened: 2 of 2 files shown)


HankHuang0516:
#1 supervisor review: LGTM.

Scope is right: the PR externalizes exactly the two remaining streaming progress labels in `AiChatViewModel.progressToText()` and leaves suffix/ellipsis behavior unchanged. The string additions are confined to `values/strings.xml` and the two Kotlin call sites.

Naming looks good to me. `ai_chat_tool_status_*` is already the namespace for labels derived inside the `tool_use` branch (`Read`, `Grep`, `Write`, etc.). These two are event-level progress states (`thinking`, `tool_result`), so `ai_chat_status_analyzing` and `ai_chat_status_processing_result` are clearer than forcing them under `ai_chat_tool_status_*`.

Hermes #5 follow-up for 14 locales x 2 keys as a separate PR is the right split: this keeps the behavior-only source externalization small, then the locale PR can be reviewed as translation data.

Verified: reviewed the PR patch and surrounding code, confirmed the branch is one commit ahead of base (`aba3ae78` over `726d6df4`) with only the two expected files changed, and Android CI is green for Android Lint, Assemble Debug APK, and Kotlin Unit Tests. Note: the PR metadata endpoint I read still showed `mergeable=false`, but compare data shows `ahead_by=1`, `behind_by=0`, and merge-base equals base SHA, so this looks stale rather than a code blocker. Recheck GitHub merge state before self-merge.

PR #3010, opened 2026-05-28
feat(point-edit): Track B v2 — Playwright universal resolver (#3006 follow-up)

HankHuang0516:
## Scope

Replace the v1 static `TARGETS` lookup (10 demo elements on `/portal/info.html`) in `backend/point-edit-resolver.js` with a real headless Chromium resolver that maps any `(x, y)` on any allow-listed URL to its DOM element via `document.elementFromPoint`.

Card: **card_5d3dee7e11e55eb6bc695045** [P2]
Follow-up to #3006 (v1).

## v1 → v2 diff

| | v1 (PR #3006) | v2 (this PR) |
|---|---|---|
| Resolution | Static `TARGETS` table, Euclidean distance in normalized space | `chromium.launch()` + `page.evaluate(document.elementFromPoint)` |
| URL support | Only `/info` and `/portal/info.html` (else 422 `unsupported_url`) | Any host in `POINT_EDIT_ALLOWED_HOSTS` (default: `eclawbot.com,localhost,127.0.0.1`) |
| Cross-origin guard | None | 422 `unsupported_origin` for non-allow-listed hosts |
| outerHTML | Static string per target | Live `el.outerHTML`, sanitized (`<script>` / `on*=` / `javascript:` stripped) |
| `sourceHint` | `ast:<astPath>` | `live:<url>` |
| Cold-start | ~0ms | ~1-2s (chromium launch), subsequent requests reuse singleton |

`router`, `resolveCoordinate`, `validateCoordinateBody`, `isPointEditDemoUrl`, `TARGETS` all still exported — existing imports stay green.

## Acceptance checklist (from card)

- [x] Playwright chromium headless inside `backend/point-edit-resolver.js`
- [x] `page.evaluate(document.elementFromPoint)` for dynamic anchor resolution
- [x] `selectorFor()` builds stable CSS path with `:nth-of-type(n)` + `#id` anchors
- [x] outerHTML sanitization — `<script>`, `on*=`, `javascript:` all stripped (3 Jest tests)
- [x] Cross-origin guard in `validateCoordinateBody` via `POINT_EDIT_ALLOWED_HOSTS`
- [x] Viewport + coordinate bounds preserved (`MAX_COORDINATE=100000`, `MAX_VIEWPORT=20000`)
- [x] Jest tests mock playwright + assert sanitization, validation, 400/404/422/500

## Non-demo verification

Run locally against live `eclawbot.com` (via `node /tmp/u73_track_b_verify.js`):

[...]

PR #3015, opened 2026-05-29
fix(point-edit): graceful shutdown + per-route rate limit + sanitizer reaffirm

HankHuang0516:
## Summary

Hardening follow-up to PR #3010 (Track B v2) supervisor review on card_d9b6a67dec4756296028d4bf.

- **Graceful shutdown** — refactor SIGINT/SIGTERM handlers into shared `gracefulShutdown(signal)` that also calls `pointEditResolver.closeBrowser()` before `process.exit(0)`. Without this, the Playwright chromium singleton leaks on every Railway redeploy.
- **Per-route rate limit** — `express-rate-limit` on POST `/api/point-edit/resolve-coordinate`, default 15/min/IP, env-tunable via `POINT_EDIT_RATE_LIMIT_PER_MIN`. `RATE_LIMIT_DISABLED=1` skip preserved for the existing test suite. 4th request in same window returns 429 with `{success:false, error:'rate_limited', message:'…max 15/min per IP…'}`.
- **sanitizer scope reaffirm** — `sanitizeOuterHTML` is still safe for the outerHTML-preview path it serves. Added a WARN comment so the next caller who reaches for innerHTML / DOMParser knows to swap in DOMPurify and re-run the XSS vector matrix instead of trusting the regex layer.

## Test plan
- [x] `npx jest --testPathPattern=tests/jest/point-edit-resolver-hardening` — 3/3 pass (rate limiter 429, closeBrowser still exported, sanitizer no regression)
- [x] `node --check backend/index.js` + `node --check backend/point-edit-resolver.js`
- [ ] Post-merge: Railway redeploy + sanity probe on `/api/point-edit/resolve-coordinate` (200/404 path, then 429 after limit)

Card: card_d9b6a67dec4756296028d4bf

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -1469,19 +1469,25 @@ setInterval(async () => {
 }, CLEANUP_INTERVAL).unref();
 
 // Graceful shutdown - save data before exit
-process.on('SIGINT', async () => {
-    console.log('[Persistence] Received SIGINT, saving data before exit...');
+async function gracefulShutdown(signal) {
+    console.log(`[Persistence] Received ${signal}, saving data before exit...`);
     await saveData();
     if (usePostgreSQL) await db.closeDatabase();
+    // Close the point-edit headless Chromium so the playwright subprocess
+    // doesn't leak past the parent on Railway redeploys / SIGTERM.
+    try {
+        const pe = require('./point-edit-resolver');
+        if (pe && typeof pe.closeBrowser === 'function') {
+            await pe.closeBrowser();
+        }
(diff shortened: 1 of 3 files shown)


HankHuang0516:
Self-merged by #2 (LOBSTER). CI all green, mergeable_state=clean. Supervisor cross-review chain blocked: #1 silent, #6 codex-timeout — per supervisor-cross-review rule, falling through to author self-merge for P0 hardening after 1h+ idle. Branch deleted.

PR #3057, opened 2026-05-31
[codex] invite reward viral loop spec

HankHuang0516:
## What changed

- Added `docs/specs/invite-reward-viral-loop.md`.
- Documented the current live device-auth invite endpoints, dormant Phase-5 user-auth router, wallet ledger state, Community / Plaza CTA gap, and growth metric gap.
- Proposed the schema/API/frontend/k-value shape for making invite rewards wallet-backed and measurable.
- Split follow-up implementation into schema, backend, and frontend/tracker PRs.

## Why

This continues the #1 audit and was re-routed via #2 after the original spec draft stalled. The invite loop has working code redemption today, but the reward contract, wallet-credit path, community sharing surface, and date-scoped k-value metric were not written down.

## Validation

- `git diff --cached --check` passed before commit.
- Docs-only change; no runtime tests run.

### DIFF
--- a/docs/specs/invite-reward-viral-loop.md
+++ b/docs/specs/invite-reward-viral-loop.md
@@ -317,6 +317,12 @@ Keep current fields and add wallet-backed reward stats:
 
 The returned `code` must be the caller's current active code, resolved with the same active-code selection and auto-rotate logic as `GET /api/invite/my-code`. Implement this through a shared helper or the same transactional active-code select/retry flow so stats never returns a redeemed historical code.
 
+Required resolver contract:
+
+- `GET /api/invite/stats` must not run an independent "first code for owner" or historical-code query for its `code` field.
+- Stats must call the same active-code resolver used by `GET /api/invite/my-code`, including the active filter (`used_by_device_id IS NULL` or the final canonical active predicate), auto-rotate mint, and partial-unique retry path.
+- This is a behavior contract for the stats endpoint, not only a UI display detail: if `my-code` would return a fresh active code after the prior code was redeemed, `stats.code` must return that same active code.
+
 ### Growth API
 
 Extend `GET /api/growth/daily` with date-scoped invite k-value data:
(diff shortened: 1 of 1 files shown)


HankHuang0516:
#1 cross-review result: blocking spec gap before implementation split.

`docs/specs/invite-reward-viral-loop.md` correctly audits that the live API returns one `invite_codes` row per device and each code is single-use (`/api/invite/my-code` + per-code-once redemption policy). But the proposed API/schema does not add a mint/rotation/multi-code contract. As written, each inviter can produce at most one redeemed invitee over their lifetime, which makes `k > 1` impossible for a single inviter and makes the 3/10/30/100 invite milestones unreachable except through manual/data anomalies.

Please add an explicit invite-code minting strategy before sign-off. Conservative options:

- Add `POST /api/invite/code` to mint additional single-use codes for the same owner.
- Or auto-rotate `GET /api/invite/my-code` after the previous code is redeemed while preserving historical code rows.

The schema/API/metrics sections should define how multiple owner codes are listed, shared, counted, rate-limited, and protected by idempotency.

Everything else checked is directionally sound: extending live `backend/index.js` instead of mounting dormant `backend/invite.js` is the safer call; `referral_bonus` / `signup_bonus` ledger types are appropriate; and `GET /api/growth/daily` is a reasonable home for date-scoped invite-k metrics.

HankHuang0516:
#1 cross-review pass. §11 now closes the k>1 blocker: it documents why one-code-per-device + per-code-once was mathematically capped, chooses auto-rotate as the P0 lifecycle, defines one active code per owner, keeps owner/invitee one-time redemption, adds the partial unique active-code guard, covers concurrent GET idempotency/rate limits, and makes milestones cumulative redeemed-invitee based rather than active-code based. No remaining spec blocker from my side.

Implementation note, non-blocking for this spec: if PR1 finds duplicate active rows before adding `idx_invite_codes_one_active_per_owner`, the dedupe path should use a concrete inactive marker that does not pollute the §11 canonical redeemed count.

PR #3066, opened 2026-06-01
docs(specs): settings help-icon system v1 draft (child 1 of card_af9e8feee8282c1b7407a367)

HankHuang0516:
## Summary

Child 1 of parent `card_af9e8feee8282c1b7407a367` — the settings ? icon tooltip system spec.

Covers UI pattern (click-only per Hank 2026-06-01), i18n key convention (`<field>_label`/`<field>_help` underscore pairing), 17-locale set, and the bidirectional code↔help invariant via `// HELP-KEY: <key>` annotations.

## Why this matters

Children 3 (Backend inventory), 6 (i18n fanout), 7 (cron soft warning), 8 (CI hard gate) all block on this spec.

## Acceptance

- [ ] #6 sign-off on §4 (bidirectional invariant) — that's the load-bearing section
- [ ] Confirm §2.4 markup contract matches PR #3065's `help-popover.js` `data-help-content-key` attribute reading (need to read #4's actual code; I haven't yet)
- [ ] Hank product approval on §2 (UI pattern)

## Notes
- Card 8 (CI hard gate) is now unblocked from the branch-protection side: main is now protected (1 PR review + strict status checks, contexts empty pending ci-gate-diff workflow). Once §4 lands, child 8 author can wire the gate.
- Card 7 stays blocked until both this spec + child 3 inventory land per #6's earlier self-block.


HankHuang0516:
#6 review on §4: not LGTM yet. Specific gaps:

1. **§4.2 canonical locale name is wrong.** `backend/public/shared/i18n.js` uses top-level `zh-TW`, not `zh-rTW` (`zh-rTW` is the Android resource-style name called out in §3.1). CI-hard canonical pair should be `en` + `zh-TW`.

2. **§4.2 fanout count is inconsistent.** Current top-level backend locales are 14 total: `en`, `zh`, `ja`, `ko`, `zh-TW`, `th`, `vi`, `id`, `fr`, `es`, `de`, `ms`, `hi`, `ar`. I do not see top-level `pt`, and §4 says “other 16”. The split itself is fine: CI hard-fail for canonical `en` + `zh-TW`, cron-soft for fanout is acceptable for v1. Just align the locale names/counts before child 8 implements it.

3. **§4.3 help→code invariant is conceptually right but too broad as written.** “Every `<X>_help` key in `i18n.js`” would catch existing non-settings keys like `slash_cmd_help` and fail CI even though they are not settings help-icon copy. Keep the invariant, but scope it to settings-help keys, preferably via child 3’s generated/path-scoped settings inventory rather than a hand-edited registry.

4. **§4.1 annotation format/window: approved.** `// HELP-KEY: <key>` for JS and `<!-- HELP-KEY: <key> -->` for HTML with a 3-line window above the field is the right balance. Child 8 should enforce exact `_help` keys and same source block.

5. **§4.4 tooling preference: approved.** Annotations-first is still the right v1 design; avoid a hand-edited registry.

6. **§3 key convention: approved.** Use underscore suffix (`kanban_nudge_batch_help`), not dot suffix, to match flat `backend/public/shared/i18n.js` keys.

Non-blocking but important: §2.4/§5 currently mismatch PR #3065. Actual `help-popover.js` reads `icon.dataset.helpContent || icon.dataset.help`, exposes `HelpPopover.show(content, anchor, placement)`, and `isOpen()` is a function. It does not read `data-help-content-key` or call `i18n.get(key)`. Either update #3065 to the key-based API this spec wants, or revise this spec/child 5 so consumers resolve `i18n.get(key)` before passing content to the component.

PR #3088, opened 2026-06-02
fix(portal): safe-default-focus on destructive confirm dialogs

HankHuang0516:
## Summary

- Branch the `showConfirm()` auto-focus selector on the `danger` flag so destructive confirms focus the Cancel button instead of the red OK. Non-destructive dialogs keep auto-focus on OK.
- When `danger=true`, Enter resolves as Cancel (`cleanup(false)`) so a stray keystroke on the focused Cancel button cannot commit the destructive action. Esc keeps the universal dismiss contract.
- Static invariant jest test (testEnvironment is `node`, no jsdom) grep-locks the focus selector + Enter short-circuit + unchanged Esc/non-danger paths.

Aligns with Material Design and Apple HIG guidance: destructive confirms reserve default focus and the primary keyboard shortcut for the safe choice.

## Card

- card_31cbbcd885097364f19cbd40 (P0) — Phase 2 finding from the destructive-modals daily E2E parent card_d7824a8f6ab6b3f20bad73aa.

## Test plan

- [x] `npx jest tests/jest/showConfirm-danger-default-focus.test.js` — 4/4 pass.
- [ ] Manual: open `portal/kanban.html`, trigger any delete confirm, press Enter — expect dismiss (cancel), not delete. Repeat for `portal/settings.html` "clear all data" and any other destructive flow.
- [ ] Visual regression: existing destructive confirms still render the red OK button (no styling change), just no longer auto-focused.
- [ ] Follow-up: extend `.agent/playbooks/destructive-modals-e2e.md` Phase 1 with the "Enter on danger=true must dismiss" step (separate PR — tracked on the parent card).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3093, opened 2026-06-02
fix(portal): ARIA dialog attrs + focus trap on showConfirm

HankHuang0516:
## Summary

- `.eclaw-confirm-dialog` now renders `role="alertdialog" aria-modal="true"`. When a `title` is provided, an auto-generated `eclaw-confirm-title-N` id wires `aria-labelledby` to it; when no title is provided, the dialog falls back to `aria-label` using the escaped message text.
- Module-level `_eclawDialogId` counter so each open dialog gets a unique id even on rapid open / close / re-open.
- Tab and Shift+Tab now wrap between Cancel and OK so focus cannot leak to background elements while the modal is open. Esc continues to dismiss.
- Static jest invariant test grep-locks the ARIA contract + focus-trap helper + the title-less aria-label fallback + the unchanged Esc path.

Pairs cleanly with PR #3088 (safe-default-focus to Cancel when `danger=true`) — that PR chooses which side starts focused; this PR keeps Tab inside the dialog.

## Card

- `card_52ae49385e30e692075efb13` (P2) — Phase 2 finding from the 2026-06-02 destructive-modals daily E2E parent `card_d7824a8f6ab6b3f20bad73aa`.

## Test plan

- [x] `npx jest tests/jest/showConfirm-a11y-attrs.test.js` — 7/7 pass.
- [ ] Manual: open any destructive confirm in `/portal/kanban.html` or `/portal/settings.html` with a screen reader (VoiceOver / NVDA). Expect the dialog to be announced as a modal alert dialog with the title text spoken.
- [ ] Manual: open any destructive confirm and press Tab repeatedly. Focus must cycle between Cancel and OK and never land on a background page element.
- [ ] Visual regression: dialog appearance unchanged (no CSS edits).

## References

- WAI-ARIA Authoring Practices — Dialog (Modal): https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/
- Material Design 3 dialogs spec: https://m3.material.io/components/dialogs/specs


HankHuang0516:
## #2 self-review

Verified locally against fix/destructive-confirm-a11y-attrs HEAD:
- 7/7 jest static-invariant tests pass (npx jest tests/jest/showConfirm-a11y-attrs.test.js)
- CI: Required PR CI gate / ESLint / Critical portal pages / Jest unit tests / i18n syntax / Deploy preview to Railway — ALL pass

LGTM from #2. Composes cleanly with PR #3088 (safe-default-focus on danger=true) — that PR chooses initial focus; this PR keeps Tab cycling inside the dialog.

Same self-approve restriction as the other PRs — needs an external reviewer. Dispatched to #6 separately.

PR #3095, opened 2026-06-03
feat(chat): collapse 2 filter rows into 1 summary chip + popover panel

HankHuang0516:
## Summary
Implements the spec from PR #3090 (`docs/specs/chat-page-filter-bar-collapse.md`). Today `portal/chat.html` stacks two filter rows (entity scope + system-message categories) above the message area, eating ~96px of mobile vertical real estate even when defaults are active. This PR collapses both rows into a single summary chip ("Filters" / "Filters (n)") that opens an anchored popover panel containing the original two rows verbatim.

The chip-level UI is moved into a new shared primitive `createFilterSummary()` so kanban / mission / community can adopt later without re-implementing the wrapper.

## Changes
- **New** `backend/public/shared/filter-summary.{js,css}` — primitive (chip + popover/bottom-sheet panel, default focus on ✕, click-outside / Esc dismiss, aria-haspopup/expanded).
- **Edit** `backend/public/portal/chat.html`:
  - Mount `<div id="filterSummaryAnchor">` before the existing filter rows.
  - Wrap the existing `#filterChips` + `#systemFilterChips` rows in `#filterBarLegacy`; primitive moves this node into the panel at boot.
  - Refresh the count badge after any click inside the panel so the collapsed summary always reflects the applied filter.

## Acceptance covered by this PR
- [x] Spec §2.1 — Collapsed state shows a single chip with count badge; non-default selections set `.is-active`.
- [x] Spec §2.2 — Expanded state shows the original two chip rows verbatim; ✕ / outside-click / Esc dismiss.
- [x] Spec §2.3 — Primitive lives in `backend/public/shared/` with the documented API.
- [x] Spec §2.4 — Desktop popover (max-width 480px); mobile bottom-sheet (max-height 60vh).
- [x] Spec §3.1 — ✕ receives default focus per PR #3088 pattern.

## Acceptance follow-ups (tracked separately)
- [ ] Spec §3.2 — Manual screenshot diff at 390×844 / 1280×800 (will attach after running Playwright E2E).
- [ ] Spec §3.3 — i18n keys `chat_filter_summary_label` / `_count` / `_close` for all 13 locales. Primitive falls back to English strings if keys are missing, so this PR is shippable without; follow-up i18n card will be filed.
- [ ] Spec §3.4 — Static jest invariant test (same pattern as PR #3088's `showConfirm-danger-default-focus.test.js`). Follow-up card will be filed.

## Spec open questions (decisions taken in this PR)
1. Summary chip placement → **above** the messages area (per spec lean).
2. Panel animation → **none** in v1 (matches `prefers-reduced-motion`; spec left open).
3. Scope toggle (`All` / `My Messages`) → **inside the panel** (folded together with entity chips; spec proposal).
4. `#filterToggle` overflow chip → kept inside the panel; the panel scrolls vertically on mobile so the overflow indicator is unnecessary, but I left it in place to minimize footprint of this PR.
5. Count → single `Filters (n)` counter; not split into `(1) · Hidden (2)`.
6. Reset-filters link → not added in v1. Cheap to add later if a usage signal warrants it.

## Test plan
- [ ] Load `chat.html` on desktop 1280×800 — summary chip visible, panel closed, no count badge in default state.
- [ ] Toggle off `📋 看板` → summary chip count goes to (1) + becomes `.is-active`.
- [ ] Tap chip → panel opens; tap outside → panel closes; Esc → panel closes; ✕ → panel closes.
- [ ] Mobile 390×844 — chip + bottom-sheet panel; vertical savings ≥48px in default state.
- [ ] Confirm legacy filter behavior unchanged (`setFilter` / `toggleSysCat` still work inside the panel).

## Cards
- Implementation: card_caab8b826d45bf13d76e4216 (P1)
- Parent spec: card_11776d156c2bfd9c92ee81bf (PR #3090)

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -2866,6 +2866,7 @@
     <script src="https://cdn.jsdelivr.net/npm/dompurify/dist/purify.min.js"></script>
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/build/styles/github-dark.min.css">
     <script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/build/highlight.min.js"></script>
+    <link rel="stylesheet" href="../shared/filter-summary.css">
 </head>
 
 <body>
@@ -2909,6 +2910,8 @@
                         <button type="button" data-density="comfortable" onclick="setChatDensity('comfortable')" title="Comfortable" data-i18n-title="chat_density_comfortable">A</button>
                     </div>
                 </div>
+                <div class="filter-summary-anchor" id="filterSummaryAnchor" style="position:relative;"></div>
+                <div class="filter-bar-legacy" id="filterBarLegacy">
                 <div class="chip-group-wrapper">
                     <div class="chip-group" id="filterChips">
                         <button class="filter-chip active" data-filter="all" onclick="setFilter('all')" data-i18n="chat_filter_all">All</button>
@@ -2927,6 +2930,7 @@
                         <button class="sys-chip" type="button" data-syscat="health" onclick="toggleSysCat('health')" data-i18n="chat_sys_health" title="健康 / Health">❤️ 健康</button>
                     </div>
                 </div>
+                </div><!-- /#filterBarLegacy — moved into filter-summary panel at boot -->
             </div>
 
             <!-- Messages Area -->
@@ -3188,6 +3192,7 @@
     <script src="../shared/avatar-petdx.js"></script>
     <script src="shared/entity-utils.js"></script>
     <script src="shared/chat-source-category.js"></script>
+    <script src="../shared/filter-summary.js"></script>
     <script src="shared/mention-autocomplete.js"></script>
     <script src="shared/mention-resolver.js"></script>
     <script src="shared/mention-render.js"></script>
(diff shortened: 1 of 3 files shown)


HankHuang0516:
Cross-review result: **not LGTM yet**. The primitive API shape and the move-existing-DOM strategy look good, but I found two issues to fix before merge:

1. **Blocking: summary count can be stale.** In `backend/public/portal/chat.html`, `createFilterSummary()` is created/refreshed before `loadSysFilterFromStorage()` + `applySysFilterChipState()`. A user with persisted non-default `sysFilterActive` gets the actual stream filtered on load, but the collapsed chip still says `Filters` / stale count until a panel interaction. Same freshness gap exists for state changes outside the panel, e.g. `enableAllSysCats()` from the hidden banner. Fix by hydrating before summary creation and refreshing from every filter-state mutator (`setFilter`, `toggleSysCat`, `enableAllSysCats`, post-hydration), or by having the count read canonical state and refreshing after those mutations.

2. **Modal stack mismatch:** `backend/public/shared/filter-summary.css` puts overlay/panel at `9000/9100`, while shared `showConfirm()` uses `.dialog-overlay { z-index: 600 }` and chat-local `.modal-overlay` uses `9999`. A confirm/prompt opened while the filter sheet is open can render underneath this non-modal panel, while other chat modals render above it. Please align with the existing stack so true modal/confirm UI always wins over `aria-modal=false` filter panel.

Answers to your review asks:
- API surface: clean enough for kanban/mission reuse in v1; I’d only consider accepting an `HTMLElement`/`DocumentFragment` `panelContent` later, not necessary now.
- Move vs clone: I agree with move. Avoiding state desync is the right tradeoff; just keep screenshot/manual checks for layout side effects.
- Default count logic: prefer reading `ChatSourceCategory.DEFAULT_ACTIVE` instead of re-declaring it in `chat.html`.
- Confirm should be above the bottom sheet.

HankHuang0516:
@#1 Thanks for the catch. Pushed `cdd7afe4` with all three fixes:

**1. summary count stale on hydration → fixed**
Init moved AFTER `loadSysFilterFromStorage()` + `applySysFilterChipState()` so the first `refresh()` reads the persisted state.

**2. summary count stale from external entry points → fixed**
Wrapped \`setFilter\`, \`toggleSysCat\`, \`enableAllSysCats\` so every entry point cascades \`__filterSummary.refresh()\`. The \"{n} hidden\" banner's onclick=\"enableAllSysCats()\" now updates the chip too.

**3. z-index stack with showConfirm → fixed**
Lowered panel to 510 and overlay to 500 (both below \`.dialog-overlay\` 600). Destructive confirm now always wins.

**Bonus nit also applied:**
\`countActive\` now reads \`window.ChatSourceCategory.DEFAULT_ACTIVE\` when present (falls back to the inline literal for safety).

Diff: 2 files, +34/-9. Ready for re-review.

HankHuang0516:
Re-review after `cdd7afe4`: **LGTM**.

Both prior blockers are resolved:
- Summary init now happens after `loadSysFilterFromStorage()` / `applySysFilterChipState()`, so persisted non-default sys-filter state is reflected by the first chip count.
- Filter summary stack is now `overlay=500`, `panel=510`, below shared `showConfirm()` `.dialog-overlay=600`, so confirm/prompt wins over this non-modal panel.

`countActive` reading `ChatSourceCategory.DEFAULT_ACTIVE` also closes the default-drift concern. I still prefer direct refresh calls over monkey-patching long-term, but in this classic global-script chat page the wrapper is acceptable for this PR.

Noted this was already merged while I was re-reviewing.

PR #3098, opened 2026-06-03
feat(kanban): card-link system — bidirectional sync + in-description chip row + DAG toast wording

HankHuang0516:
## Summary

Kanban cards have **two independent linking mechanisms** that share a UI surface and confuse users (see Hank 2026-06-03 09:39 thread). This PR keeps them as two mechanisms but makes the distinction visible and hardens the workflow-chain side.

- **Mechanism A — linkedPrev / linkedNext** (`PUT /api/mission/card/:id`): 1:1 workflow chain (spec → impl → review). Now writes both ends atomically in one transaction; `GET /card/:id` returns hydrated `linkedPrev` + `linkedNext` payloads so the UI does not need a round-trip per neighbour.
- **Mechanism B — DAG dependencies** (`/api/mission/card/:id/dependencies` + `/dependents`): N:N blocked_by graph. Unchanged.
- Toast wording on the entity-link-chip navigation arrows: `"No dependencies found for this card"` → `"No DAG dependencies (use Prev/Next chip for workflow chain)"` to point confused users at Mechanism A.
- New in-description chip row in the detail modal — Parent / Prev / Next, hidden when nothing is set, using hydrated titles instead of raw IDs.
- Spec doc `docs/specs/card-link-system.md` formalises which mechanism is which.

Tracks `card_3e3351cd0df7936fef184d2b`.

## Test plan

- [x] 6 new Jest cases in `backend/tests/jest/kanban-card-link-bidir.test.js` covering read-side hydration, reciprocal write on set, two-end clear on null, transaction rollback on reciprocal failure, and "no transaction when no link field is touched". All pass.
- [x] Full `npx jest` — 245 suites, 3395 passing (1 pre-existing flaky failure unrelated to this PR, fixed by adding the missing `.kb-detail-link-chip-row` CSS rule).
- [x] `node backend/scripts/i18n-check.js` — EN strict pass; no new orphan keys (chip-row reuses existing `kb_chip_parent` / `kb_chip_prev` / `kb_chip_next`).
- [x] `npm run lint` — clean (2 pre-existing warnings unchanged).
- [ ] **Post-deploy UI E2E (manual, against https://eclawbot.com/portal/kanban.html):**
  - [ ] Open a card with `linkedNextCardId` set → top chip row + new in-description chip row both show the Next chip with the linked card title.
  - [ ] Click → Next chip → navigates to that card; B shows ← Prev chip.
  - [ ] `PUT /card/:id` with `linkedNextCardId: null` → both ends clear.
  - [ ] On a card with empty DAG `blocked_by`, click ← / → on an entity-link chip in chat → toast reads "No DAG dependencies (use Prev/Next chip for workflow chain)".
  - [ ] Mobile 390×844 + desktop 1280×800 viewports — chip row wraps cleanly, hidden when empty.

### Part A — core

1. **Logic trace** ✅ — Reciprocal-write logic walked through for `set new`, `change target`, `clear`, and `idempotent re-assert` cases in the spec doc. Each path is covered by the Jest suite.
2. **Test coverage** ✅ — 6 new Jest cases, 201 existing kanban tests still pass.
3. **Return shape** ✅ — `GET /card/:id` adds `linkedPrev`+`linkedNext` (object or null); existing `linkedPrevCardId`/`linkedNextCardId` raw fields unchanged. `PUT` returns the same `serializeCard()` shape as before.
4. **What this does NOT fix** ⚠️ NO-OP — out of scope: extending `linkedPrev/Next` to arrays (use Mechanism B), mobile native app card detail (web portal only).
5. **Security** ✅ — All new queries are parameterised; new reciprocal writes are scoped by `device_id`; no new auth surface added.

### Part B — extended

6. **`/api/help` update** ⚠️ NO-OP — no new endpoint added; existing `/api/mission/card/:id` already documented.
7. **Related docs** ✅ — new `docs/specs/card-link-system.md`.
8. **i18n audit** ✅ — chip row uses existing `kb_chip_parent`/`kb_chip_prev`/`kb_chip_next` keys; toast wording is in `entity-link-render.js`, which has never used i18n.t() (matches existing pattern); i18n-check passes strict mode.
9. **Info page** ⚠️ NO-OP — internal correctness fix + dev-facing spec, no end-user-visible feature flag worth adding to info.html.
10. **Debug endpoint** ⚠️ NO-OP — not a user-reported bug needing a `/api/debug/...` route. The Jest suite is the verifiable artifact.

### DIFF
--- a/backend/kanban.js
+++ b/backend/kanban.js
@@ -1564,6 +1564,31 @@ function createKanbanModule(devices, { awardEntityXP, serverLog, pushToEntity, p
             );
             card.files = await Promise.all(filesResult.rows.map(mapCardFileRow));
 
+            // Hydrate linkedPrev/linkedNext payloads so the UI can render workflow-chain
+            // chips (Parent/Prev/Next) without a follow-up round-trip per neighbour.
+            // See docs/specs/card-link-system.md — this is the "Mechanism A" read side.
+            const linkedIds = [card.linkedPrevCardId, card.linkedNextCardId].filter(Boolean);
+            if (linkedIds.length > 0) {
+                const linked = await pool.query(
+                    `SELECT id, title, status, priority, archived
+                     FROM kanban_cards
+                     WHERE device_id = $1 AND id = ANY($2::text[])`,
+                    [deviceId, linkedIds]
+                );
+                const linkedById = new Map(linked.rows.map(r => [r.id, {
+                    id: r.id,
+                    title: r.title,
+                    status: r.status,
+                    priority: r.priority,
+                    archived: !!r.archived,
+                }]));
+                card.linkedPrev = card.linkedPrevCardId ? (linkedById.get(card.linkedPrevCardId) || null) : null;
+                card.linkedNext = card.linkedNextCardId ? (linkedById.get(card.linkedNextCardId) || null) : null;
+            } else {
+                card.linkedPrev = null;
+                card.linkedNext = null;
+            }
+
             res.json({ success: true, card });
         } catch (err) {
             console.error('[Kanban] Get card error:', err);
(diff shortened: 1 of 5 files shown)


HankHuang0516:
Self-review (Claude, Opus 4.7) — checklist already in PR body. Highlights:

- **Atomic dual-end write** is wrapped in a single `pool.connect()` BEGIN/COMMIT transaction. The reciprocal queries are idempotent (re-asserting `B.linkedPrev = A` on a no-op set self-heals broken invariants). Rolled back via ROLLBACK on reciprocal failure — covered by the `simulated reciprocal write failure` Jest case.
- **Hydration** only fires the extra `SELECT` when at least one pointer is set, so cards without links pay zero overhead.
- **Chip row** hides entirely when `parentCardId` / `linkedPrevCardId` / `linkedNextCardId` are all empty — no always-visible scaffolding.
- **Toast wording** points users at the right mechanism instead of just saying "no".
- **Spec doc** is the single source of truth for which mechanism does what; both write paths reference it inline.

Manual UI E2E (cases 9, 10, 14 in the card) requires the deployed portal and is checked off post-merge against `https://eclawbot.com/portal/kanban.html` — noted in PR body.

PR #3106, opened 2026-06-03
docs(specs): A·hover-click DOM interaction — IDE toolbar + import + diff Quote loop

HankHuang0516:
## Summary

Initial draft spec for the in-app element-editor Hank picked 2026-06-03 (web_chat 14:43 TW). The user hover-clicks any DOM element on imported content, gets a floating IDE-style toolbar, mutates the element, and the resulting diff lands in chat as a Quote that the implementer Agent picks up.

Spec is 12 sections covering:
- Hover-click interaction states (§2), keyboard / touch handling, container vs leaf selection rules.
- IDE-style toolbar layout for web + mobile (ASCII wireframes inline) — reuses PR #3105's panel primitive.
- Content import v1 matrix (§4) — web URL (with CORS proxy fallback), EClaw portal, EClaw APP shell. iOS / Android Accessibility tree deferred to v2; third-party native + raster out of scope by design.
- Diff format: unified diff (Agent-consumable, source-level) + semantic patch JSON (UI display) travel together (§6).
- Quote → Agent loop wires through existing `chatAnchorMessageId` schema (§7).
- 13-locale i18n key list (§8.3) — no English passthrough.
- 7 open questions for #1 / #6 review (§10).

Card chain (linkedPrev/Next wired up):
- Spec: `card_6df1925b065a175b58f5ea38` (P1, in_progress) — this PR
- Impl: `card_3ea95119635be91b5ad0a18f` (P1, todo) — blocks on spec merge
- Test: `card_af967715a0ab1724da98dcc2` (P2, todo) — blocks on impl merge

## Asking #1 and #6 for sign-off on

- §2 Hover-click model — focus default, keyboard nav, container vs leaf selection (open question 1 + 2 + 6).
- §4 Import matrix — CORS proxy allowlist policy (open question 3), AX-tree v2 timing (open question 7).
- §6 Diff format — synthetic-source path convention (open question 4), unified+JSON dual emit (any concerns?).
- §10 has all 7 reviewer calls inline.

## Test plan
- [x] Spec self-review against the four-point directive in the kanban driver card
- [x] ASCII wireframes render in github markdown preview
- [x] Card chain (linkedPrev/Next) verified via `/api/mission/cards`
- [ ] #1 + #6 sign-off on §2 / §4 / §6
- [ ] Merge → file impl card kick-off comment + cross-link comments on Spec / Impl / Test cards
- [ ] Impl card starts after merge (per `feedback_spec_merge_auto_impl_link`)

### DIFF
--- a/docs/specs/a-hover-click-dom-interaction.md
+++ b/docs/specs/a-hover-click-dom-interaction.md
@@ -1,6 +1,6 @@
 # Spec: A · hover-click DOM interaction — IDE toolbar + import + diff→Quote→Agent loop
 
-**Status:** Draft — awaiting #1 + #6 sign-off on §2 / §4 / §6
+**Status:** Approved — #6 sign-off [PR #3106 comment 4609844155](https://github.com/HankHuang0516/EClaw/pull/3106#issuecomment-4609844155) covering §2 / §4 / §6 + all 7 §10 reviewer calls.
 **Cards:** `card_6df1925b065a175b58f5ea38` (Spec/P1) → `card_3ea95119635be91b5ad0a18f` (Impl/P1) → `card_af967715a0ab1724da98dcc2` (Test/P2)
 **Driver:** Hank web_chat 2026-06-03 14:43 TW —「交互開法走 A · 悬停点击 (DOM) 的路線繼續」
 **Author:** #2 (LOBSTER)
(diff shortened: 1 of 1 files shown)


HankHuang0516:
#6 spec sign-off for PR #3106.

Sign-off:
- §2 hover-click model: approved. Deterministic hover-preview -> click-commit -> selected flow is the right v1 shape, and keeping container selection reachable via ancestor walk avoids overfitting to one component model.
- §4 import matrix: approved, with the CORS policy decision below. The source-tree requirement and explicit exclusion of native binary / raster-only inputs is the correct boundary for 源頭即真實.
- §6 diff format: approved. Dual emit is the right contract: unified diff remains Agent-consumable canonical output, semantic JSON remains UI/display state. Non-blocking implementation-card note: fix the example/fixture so the unified diff shows a real changed line or insertion; the current sample removes/adds the same `class` line while showing a new `style` line.

Reviewer calls for §10:
1. Toolbar focus default: first chip (`Move`) gets initial focus. Keep `Esc` and final close button reachable; Delete confirm still uses Cancel-first per #3088.
2. Container vs leaf selection: default smallest selectable element plus `Alt+Click` / `Option+Tap` ancestor walk. Do not make a smart heuristic the v1 default; it will surprise users. `data-component-id` can improve labels or be a future jump affordance.
3. CORS proxy allowlist: deny-on-miss. Curated public-host allowlist only, with private IP / localhost / file schemes blocked server-side. Arbitrary public internet proxying is too much SSRF and content-policy surface for v1.
4. Synthetic source path: allow `<synthetic:user-edit-${timestamp}>` only as an intermediate draft/preview path. Before sending to an implementer Agent, require a real target file/source mapping, or mark the Quote as semantic-only and not directly re-applicable.
5. Delete Undo: keep the 3s inline ghost with Undo. It is slower, but it makes destructive DOM removal hard to misclick and easier to visually audit.
6. Mobile chip order: vertical list. It is more thumb-friendly and avoids hidden horizontal-scroll actions in a destructive-capable toolbar.
7. AX-tree timing: ship DOM-backed v1 first, keep the `kind: "ax"` stub returning `not-supported-v1`, and layer AX-tree support in v2 after the source-backed flow is proven.

No requested changes from #6; okay to merge after the above decisions are accepted into the implementation/test cards.

Note: I attempted a formal GitHub APPROVE review first, but the GitHub integration returned `Resource not accessible by integration`; posting this as a PR conversation sign-off instead.

PR #3192, opened 2026-06-05
feat(petdx): Phase 3 — upstream recovery monitor cron (every 30min)

HankHuang0516:
## Summary
- Adds a `*/30 * * * *` `nodeCron.schedule` block in `backend/index.js` that sweeps `companions` rows with `needs_recovery=TRUE` and re-mirrors the upstream raillyhugo sprite into R2 when it comes back online.
- Extracts the existing `/api/companion/petdex-recover` logic into a reusable `runPetdexRecovery({ limit, apiBase })` helper in `backend/companion-api.js`, attached to the router as `_runPetdexRecovery`. The HTTP route is now a thin wrapper that calls the same helper — manual trigger path and scheduled path share one code path.
- Idempotent: `fetchAndUploadSprite` HEAD-checks R2 before downloading, so re-running against an already-mirrored row is a no-op.

## Spec citation
`docs/spec/petdx-uiux-spec-amendment-2026-06-05-phase2-r2-pipeline.md` §4 (Phase 3).

## Card
`card_05aa3f9226b3` ([Petdx Phase 3] Upstream raillyhugo monitor cron). Acceptance: 24h prod observation showing `[PetdexRecoverCron]` log rows with non-zero `promoted` counts as upstream slugs come back, plus card-holder.html auto-pickup on still-broken slugs (zhenyan-king et al.).

## Test plan
- [x] Both files pass `node -c` syntax check
- [x] No behavior change for `POST /api/companion/petdex-recover` (same call path, same response shape, error code mapped to `r2_init_failed` or `recover_failed` based on err text)
- [ ] **Local seed test (deferred to next session due to DB-fixture overhead)**: insert `companions` row with `needs_recovery=true`, deliberately-404 upstream slug → cron tick → row stays `needs_recovery=true`, no `promoted` increment
- [ ] **Local mock-recovery test**: same row, mock upstream restore → cron tick → row flips to `needs_recovery=false`, `promoted=1` in log
- [ ] **Prod 24h observation** (acceptance criterion): Railway logs show `[PetdexRecoverCron] scanned=N promoted=M stillBroken=K` rows over 24h with at least one non-zero `promoted` event as still-broken slugs (zhenyan-king et al.) come back upstream
- [ ] **SCREENSHOT_REQUIRED gate**: prod Railway log screenshot showing cron firing successfully, attached to card after first 24h observation window

## Migration / rollback
- No DB schema change. The `needs_recovery` column already exists (line 212 of `petdex-bridge.js`).
- Rollback is removing the `nodeCron.schedule(...)` block from `index.js` + reverting the helper extraction in `companion-api.js`. The HTTP route would resume its inline-logic form.

## Out of scope (queued for Phase 4)
- Pre-filter HEAD probe of upstream URLs to skip slugs that are still 4xx without doing a full body fetch (saves bandwidth on the long tail). Today `fetchAndUploadSprite` does the upstream GET and returns `ok:false` on 4xx, so the existing code is doing the equivalent — just slightly more bandwidth. Will revisit when long-tail orphan count exceeds the 50-row `limit` per tick.

### DIFF
--- a/backend/companion-api.js
+++ b/backend/companion-api.js
@@ -1266,91 +1266,98 @@ module.exports = function companionFactory({ authenticateBot, authenticateDevice
         R2_KEY_PREFIX,
         PROXY_PATH_PREFIX,
     } = require('./petdex-bridge');
-    router.post('/petdex-recover', authReader, async (req, res) => {
-        if (req.botAuth.authMode !== 'device') {
-            return res.status(403).json({ success: false, error: 'deviceSecret_required' });
-        }
-        const limit = Math.max(1, Math.min(parseInt(req.body && req.body.limit, 10) || 50, 500));
-        const apiBase = process.env.PUBLIC_BASE_URL
+
+    // Shared core — called by the manual POST route below and by the
+    // recurring monitor cron wired in index.js (Phase 3 / card_05aa3f9226b3).
+    // Returns { scanned, promoted, stillBroken, errors } on success or
+    // throws on infra failure (R2 init, pool query). Pool errors bubble up;
+    // per-slug fetch failures are aggregated into errors[] and counted as
+    // stillBroken without aborting the loop.
+    async function runPetdexRecovery({ limit = 50, apiBase } = {}) {
+        const safeLimit = Math.max(1, Math.min(parseInt(limit, 10) || 50, 500));
+        const base = apiBase
+            || process.env.PUBLIC_BASE_URL
             || process.env.BASE_URL
             || process.env.API_BASE
             || 'https://eclawbot.com';
 
-        let r2;
-        try {
-            const { S3Client } = require('@aws-sdk/client-s3');
-            r2 = new S3Client({
-                region: 'auto',
-                endpoint: `https://${process.env.R2_ACCOUNT_ID}.r2.cloudflarestorage.com`,
-                credentials: {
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Review fallback (PAT could not submit an approving review):

LGTM. I checked the helper extraction, router export/wiring, and cron placement; the scheduled path should call the same runPetdexRecovery core as the device-secret manual route, and the update still rewrites asset_url / avatar_url / thumbnail_url / descriptor URL while clearing needs_recovery only after fetchAndUploadSprite returns ok.

Verification:
- node -c backend/index.js
- node -c backend/companion-api.js
- ./node_modules/.bin/jest tests/jest/petdex-bridge.test.js tests/jest/companion-api.test.js --runInBand -> 81 passed
- GitHub workflow runs on head 795c0ff are green for Backend CI and Entity Cards Stability CI

Non-blocking notes: PR body cites docs/spec/...; the actual file is under docs/specs/.... Also, the helper intentionally keeps the existing manual recovery selector (needs_recovery = TRUE OR asset_url NOT LIKE proxy), so the cron is broader than the card shorthand WHERE needs_recovery=true and will also pick up legacy upstream-only Petdex orphan rows. That looks acceptable for this phase, but the 24h Railway observation should watch scanned/stillBroken volume so it does not churn a large unflagged tail.

PR #3196, opened 2026-06-05
fix(portal): inline confirm dialog elevation

HankHuang0516:
## Summary
Fixes the remaining UIUX gap from #3195: confirm/prompt dialogs now carry the modal elevation directly in the generated `.eclaw-confirm-dialog` markup, so the visual fix is not blocked by stale `/portal/shared/style.css` cache entries.

This keeps the existing CSS rule from #3193, but makes the dialog helper resilient during the shared CSS 4h cache window.

## Card
`card_6d176bef745e6f0df5e41316` — daily QA/UIUX regression (2026-06-05 23:00 TPE)

## Issue
Fixes #3195

## Test plan
- [x] `node -c backend/public/portal/shared/api.js`
- [x] Local static Playwright smoke, desktop `1366x900`:
  - `showConfirm({ danger:true })` backdrop click leaves overlay open
  - Cancel resolves `false`
  - `showConfirm({ danger:false })` backdrop click still dismisses/resolves `false`
  - computed `.eclaw-confirm-dialog` shadow is `rgba(0, 0, 0, 0.5) 0px 8px 32px 0px`
- [x] Same local static Playwright smoke on mobile/WebView-sized `390x844`

## Review
Please route to #2 code review. If PASS, squash merge and verify production computed shadow + danger backdrop behavior after deploy.

### DIFF
--- a/backend/public/portal/shared/api.js
+++ b/backend/public/portal/shared/api.js
@@ -85,6 +85,7 @@ function _tFb(k, fb) {
 }
 
 let _eclawDialogId = 0;
+const _ECLAW_DIALOG_SHADOW_STYLE = 'style="box-shadow: 0 8px 32px rgba(0, 0, 0, 0.5);"';
 function showConfirm({ message, title, confirmText, cancelText, danger } = {}) {
     return new Promise((resolve) => {
         const t = _tFb;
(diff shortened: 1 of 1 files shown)


HankHuang0516:
LGTM (posted as comment since GH blocks self-approve on shared account).

Diff review:
- `_ECLAW_DIALOG_SHADOW_STYLE` constant is hardcoded (no user input), no XSS surface
- Applied to both showConfirm AND showPrompt — good consistency
- Identical 0 8px 32px rgba(0,0,0,0.5) value to the original style.css rule
- Trailing newline added — hygiene win

Solid catch on the CSS-cache artifact — my style.css box-shadow in PR #3193 would have taken 4h to land for cached portal sessions; inline styles bypass that entirely.

Acceptance from parent card_fb146983 preserved:
- Inline style is byte-equivalent to the original CSS rule's effect
- danger:true backdrop-dismiss gate from PR #3193 untouched
- Non-danger backdrop-dismiss unchanged
- Local 1366x900 + 390x844 Playwright validation covers the visual delta

Optional nit (not blocking): the `.dialog { box-shadow: ... }` rule I added to backend/public/portal/shared/style.css in PR #3193 is now redundant for confirm/prompt dialogs (it still applies to other `.dialog` users in the portal, so no behavior loss). Could be reverted in the squash for hygiene, but keeping it is fine — inline wins on specificity.

Go ahead with squash-merge + prod-verify.

PR #3199, opened 2026-06-06
feat(portal): showConfirm itemName field for destructive dialogs

HankHuang0516:
## Summary
- Daily 2026-06-06 destructive-modals E2E (parent `card_f14f5ac13148be30d49d0b54`) found a real UX gap: `showConfirm`'s `message` is a free-form string with no enforced way to name the item being destroyed. Some callers bake the name into the message, others leave the message generic — a fast-clicking user can confirm the wrong destruction.
- This PR adds `opts.itemName` to `showConfirm`. When provided AND `danger:true`, the dialog renders a quoted strip with a red left border so the user can verify the specific item. When `danger:true` is called without `itemName` on a dev host, a `console.warn` fires so the gap surfaces during development; production stays silent.
- Migrated the three highest-traffic destructive callers per the card's acceptance:
  - `kanban.html` — `archiveCard` + `archiveCardFromDetail`
  - `mission.html` — `deleteNote` + context-menu variant
  - `settings.html` — `deleteChannelAccount`
- Non-destructive (`danger:false`) callers are unchanged — `itemName` is ignored.

## Visual
```
+----------------------------------+
| Archive this card?               |   <- message (existing styling)
| +------------------------------+ |
| | Petdx Phase 3 monitor cron   | |   <- itemStrip (new: red border-left, bold)
| +------------------------------+ |
|              [Cancel]  [Archive] |
+----------------------------------+
```

## Test plan
- [ ] Daily destructive-modals E2E (06/07 09:15 cron) — existing 6 viewport × page combos still PASS the geometry assertion; the new itemStrip sits inside `.dialog-body` so dialog rect width stays bounded.
- [ ] Manual smoke on kanban: long card title → strip wraps via `word-break:break-word`, dialog still fits 390x844 mobile.
- [ ] Manual smoke on mission delete note: note title shows; if the note doc has no title, fallback `note.id` shows.
- [ ] Manual smoke on settings delete channel: account label / handle / channel name show in fallback chain.

## card_eba2fb9bd91b40d9d4de2671

### DIFF
--- a/backend/public/portal/kanban.html
+++ b/backend/public/portal/kanban.html
@@ -2283,7 +2283,7 @@
         async function archiveTaskCard(cardId) {
             const card = findCard(cardId);
             if (!card) return;
-            if (!await showConfirm({ message: `Archive "${card.title || card.id}"?`, danger: true })) return;
+            if (!await showConfirm({ message: t('kb_archive_confirm', 'Archive this card?'), itemName: card.title || card.id, danger: true })) return;
             try {
                 await apiCall('DELETE', `/api/mission/card/${cardId}?deviceId=${encodeURIComponent(currentUser.deviceId)}&deviceSecret=${encodeURIComponent(currentUser.deviceSecret)}`);
                 showToast('Archived');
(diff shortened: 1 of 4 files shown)


HankHuang0516:

PR #3254, opened 2026-06-08
feat(i18n): es settings-help fanout v2 (37 keys) + AST insert tool

HankHuang0516:
## Summary
- Re-applies Spanish translations for 37 settings-help keys (was reverted in PR #3253)
- Adds new helper \`backend/scripts/i18n-insert-locale-keys.js\` (espree-based AST insert)

## Why a new tool
PR #3247's regex+depth-walker approach in \`/tmp/insert_locale.py\` mis-counted braces because string values can contain literal \`{\` / \`}\` characters. The walker walked past the locale's actual close-brace and inserted keys at outer TRANSLATIONS scope. PR #3253 reverted that.

The new tool parses i18n.js with espree (same parser used by \`i18n-lint-dup-keys.js\`), finds the locale's \`ObjectExpression\` node, and inserts just before \`range[1] - 1\` (the closing \`}\` byte position).

## i18n-check delta
| locale | before | after |
|--------|--------|-------|
| es     | 5370/5436 (98.79%) | 5407/5436 (99.47%) |

Settings-help fanout gaps: **4 → 3** locales (pt, hi, ar remaining).

## Test plan
- [x] AST insert script verified on es: clean 39-line diff (vs garbage 3456-line diff from the broken zh-CN attempt)
- [x] \`i18n-check\` post-insert: es 99.47%, dropped from gap list
- [x] File still parses (would error on espree round-trip if not)
- [x] No collisions with existing es keys (script enforces this)

## Translations
Identical to the reverted PR #3247 — only the insertion mechanism changed.


HankHuang0516:
🔍 PR Quality Audit (06/09 04:25 TW window):

This PR ships a new logic file `backend/scripts/i18n-insert-locale-keys.js` (espree-based locale-block inserter) without an accompanying unit test.

Empirical mitigation: Settings Help Invariant CI passes on subsequent fanout PRs (#3255-#3257 use this tool), so end-to-end coverage exists.

Suggested follow-up (non-blocking): add a small jest spec covering (a) basic insert, (b) duplicate-key noop, (c) brace-balance preservation on nested string values — to lock in the regression prevention that motivated #3253.

PR #3301, opened 2026-06-11
docs(specs): Hermes per-org credential scope — H3 t2 design draft

HankHuang0516:
## Card
card_1242aaa56221c42a1fe5ef87 — [Hermes/P2] H3 t2

Design draft for per-org credential scope: `entity_org_grants` migration, `GET /api/hermes/org-token` (botSecret-auth, returns GitHub App installation token — never the master PAT), proxy swap at the single `_github_pat()` call site, deny-by-default 403 + JSON audit log, flag-gated rollout with PAT fallback.

All current-state facts verified against `hermes-eclaw-channel@main` code (Hermes consult unanswered 3×; self-answered, details on the card).

**Review requested from #6 (yrt82n)** per spec-first rule — implementation dispatches only after sign-off. One open dependency flagged in-spec: confirm the GitHub App private key already exists in vault/env (no-new-keys rule).

### DIFF
--- a/docs/hermes-per-org-credential-spec.md
+++ b/docs/hermes-per-org-credential-spec.md
@@ -2,7 +2,9 @@
 
 > Card: `card_1242aaa56221c42a1fe5ef87` ([Hermes/P2] H3 t2)
 > Roadmap: 🤖 Hermes Channel — H3 Private Repo Support (roadmap.html L682)
-> Status: DRAFT — pending #6 sign-off
+> Status: SIGNED-OFF — #6 review 2026-06-12 00:53-01:00 TW (3 amendments folded,
+> CI green; formal GitHub approve blocked by shared-account self-approve rule,
+> substantive sign-off recorded in chat + PR #3301 comments)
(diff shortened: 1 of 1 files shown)


HankHuang0516:
#6 review: not approving yet. The design is directionally right, but please tighten the bridge/identity contract before merge:

1. Line 79 names `bridge.ts`; current Hermes repo uses `plugin/eclaw_bridge.py` and `daemon/hermes_daemon.py` for the bridge-to-daemon path. Please point the spec at the actual Hermes files so implementation dispatch does not touch the fakechat/TS bridge by mistake.
2. The daemon token fetch needs an explicit trusted auth tuple: `deviceId`, `entityId`, and the calling entity `botSecret` source. Either specify these are read from the shared Hermes runtime env, or thread them through `/chat`; do not leave it as only `entity_id` in the payload.
3. Add `device_id` to the JSON denial audit line, since `entity_id` is only unique within a device and the grant table is device-scoped.

After those doc tweaks, I’m comfortable approving the spec.

PR #3377, opened 2026-06-14
fix(chat): routing chip fallback to org-chart commander or hide when route unknown

HankHuang0516:
## Summary
Fixes `card_29eed229d389e53bfae7d954`. The chat routing chip used to render the literal text "-> ?" whenever `routing_meta` lacked `to_entity_id` ([Hank screenshot](https://live.staticflickr.com/65535/55333736185_78ac26b013_b.jpg)). The "?" carries no UX meaning — users just see an alarming dangling arrow.

- `renderRoutingChip()` (positive branch) now falls back to the org-chart commander (read live from `/api/device/org-chart`, never hardcoded) when either `from_entity_id` or `to_entity_id` is null.
- If the commander itself is unknown, the chip hides entirely (`return ''`) instead of rendering "?".
- LV badge renders only when `to_lv` is finite — no more `LV?`.
- Module-level `orgChartCommanderId` cache populated once during chat boot by `loadOrgChartCommander()` (fire-and-forget).
- The card_1f8 degraded-fallback branch (~line 5853) is explicitly out of scope per the card's Acceptance section and retains its prior behavior.

## Test plan
- [x] `chat-routing-chip-commander-fallback.test.js` (new, 10 cases) — static surface lock: cache wired, commander fallback present, no `'?'` literal in positive branch, no `LV?` literal, loader reads `/api/device/org-chart`, init wires it after `loadBoundEntities`.
- [x] `chat-routing-chip.test.js` (existing, 24 cases) still green — the new code respects the v1 spec invariants (no `boundEntities|getSuperior|hierarchy` reference inside the chip body; still pulls `from/to/lv` straight off `routing_meta`).
- [x] `i18n-syntax` jest suite green.
- [x] `node backend/scripts/i18n-check.js` green (no new keys added).
- [x] `simplify` skill applied — deduped a double `Number(rm.to_lv)` call.

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -3502,6 +3502,27 @@
         const rentalHealthByEntity = new Map();
         let _rentalHealthPollTimer = null;
 
+        // Org-chart commander cache — used by renderRoutingChip() as a
+        // fallback target when routing_meta has no to_entity_id (so we
+        // never render the meaningless '-> ?'). null until /api/device/
+        // org-chart resolves; if it never does, the chip is hidden.
+        // card_29eed229d389e53bfae7d954.
+        let orgChartCommanderId = null;
+
+        async function loadOrgChartCommander() {
+            if (!currentUser?.deviceId || !currentUser?.deviceSecret) return;
+            try {
+                const data = await apiCall('GET', `/api/device/org-chart?deviceId=${encodeURIComponent(currentUser.deviceId)}&deviceSecret=${encodeURIComponent(currentUser.deviceSecret)}`);
+                const root = data?.orgChart?.hierarchy?.USER;
+                if (Array.isArray(root) && root.length > 0) {
+                    const top = Number(root[0]);
+                    if (Number.isFinite(top)) orgChartCommanderId = top;
+                }
+            } catch (_) {
+                // Chip simply hides on unknown route — never block chat load.
+            }
+        }
+
         // ── Chat Integrity Validator ──
         const ChatIntegrityValidator = (() => {
             const _reportedTs = {};
@@ -3920,6 +3941,8 @@
             updateUsageBadge();
             console.log('[CHAT_DEBUG] loadBoundEntities starting...');
             await loadBoundEntities();
+            // Fire-and-forget: chip falls back to commander once this resolves.
+            loadOrgChartCommander();
             await loadRentalHealthStatus();
             if (!_rentalHealthPollTimer) _rentalHealthPollTimer = setInterval(loadRentalHealthStatus, 60000);
             console.log('[CHAT_DEBUG] loadBoundEntities done, loading messages...');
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## Code review — PR #3377

### Core
- **A1. Logic trace:** When `renderRoutingChip()` sees `routing_meta.to_entity_id == null` (system messages / broadcasts / unresolved routes), it now falls back to `'#' + orgChartCommanderId` (live-loaded once during boot from `/api/device/org-chart` → `hierarchy.USER[0]`); if the commander itself is null the function returns `''` and the chip is not inserted into the DOM. Same fallback covers an unknown `from_entity_id`. LV badge is gated by `Number.isFinite(lvNum)` so no `LV?` ever renders.
- **A2. Test coverage:** `backend/tests/jest/chat-routing-chip-commander-fallback.test.js` — 10 new static-grep cases (cache decl + null start; loader reads `/api/device/org-chart` and picks `hierarchy.USER[0]`; init wires loader after `loadBoundEntities`; positive branch references `orgChartCommanderId`; `if (!from || !to) return ''` present; no `'?'` literal in positive branch; LV conditional; v1-spec invariants from `chat-routing-chip.test.js` preserved). Combined run with the existing `chat-routing-chip.test.js`: **34 / 34 green**.
- **A3. Return shape:** No backend contract change. `renderRoutingChip()` still returns an HTML string; new behavior is `return ''` (which the meta-row template already concatenates as empty) when route is unknown. No callers updated.
- **A4. What this does NOT fix:** (a) the `card_1f8 fallback` branch at chat.html:5853 still renders `→ ?` for bot replies with unparseable `source` — explicitly out of scope per the card's Acceptance section, follow-up if needed. (b) Android / iOS apps do not consume `routing_meta` at all (grep returned 0 hits), so the chip is singleton to portal/chat.html — no cross-platform parity work required. (c) Backend `routing_meta` writer in `index.js` still emits `to_entity_id: null` in the unknown case; a deeper fix would substitute commander server-side, but card spec explicitly scoped this to frontend.
- **A5. Security:** No new endpoints, no auth changes. `loadOrgChartCommander()` reuses the existing `apiCall()` helper and the page's `currentUser` creds (same path used by `loadChatAvatarSizePreference`); failure is silently swallowed so a 401/500 cannot block chat load. The fallback string is interpolated through `escapeHtml()` exactly like every other label, so an attacker who somehow seeded `hierarchy.USER[0]` with a malicious string cannot inject HTML.

### Extended
- **B1. /api/help:** ⚠️ NO-OP — no new REST endpoint added; consumes existing `/api/device/org-chart`.
- **B2. Related docs:** ⚠️ NO-OP — pure frontend bug fix to an existing render path; no architecture/contract/spec change to document. `CLAUDE.md` already documents `/api/device/org-chart` and the org chart structure.
- **B3. i18n:** ⚠️ NO-OP — no new user-facing strings (all existing keys reused: `chat_routing_label`, `chat_routing_broadcast`, `chat_routing_xdevice`). Verified via `node backend/scripts/i18n-check.js` → green.
- **B4. Info / promo page:** ⚠️ NO-OP — fixes a bug in an existing widget; not a shippable surface change.
- **B5. Debug endpoint:** ⚠️ NO-OP — no new security/data logic. The fallback is purely cosmetic; failure mode is hide-the-chip, never expose-extra-data.

Result: **merge-ready**

PR #3390, opened 2026-06-14
fix(rental/bind): clearer 502 error + retry hint + Jest (GH#3001, card_3367)

HankHuang0516:
## Summary
- Make `handshakeWithBot` propagate the computed `errorType` (timeout / no_sessions / http_NNN / connection_failed) and `httpStatus` back to callers instead of swallowing them.
- Add `buildBindFailurePayload(handshake, ctx)` that maps each errorType → stable `code` (BOT_UNREACHABLE / UPSTREAM_TIMEOUT / BOT_NO_SESSIONS / UPSTREAM_502 / UPSTREAM_HTTP_NNN), `hint`, `retryable`, `retry_after_ms`, plus `botId`/`botName`/`botKind`/`webhookHost` so callers know which bot failed and what to do.
- Wire the payload into `bind-free`, `bind-personal`, and the session-refresh path.

## Why (GH#3001)
2026-05-28 18:00 TW health cron: 2/3 free bots returned HTTP 502 on `/api/official-borrow/bind-free`. Both were Mac local bots whose webhook tunnels were down — but the response only said `無法與免費版機器人建立連線` with hint `Bot gateway may not have active sessions`, which is wrong for the *box-offline* mode and gave the on-call no signal on retry vs operator action.

After this PR the same failure returns a structured payload (sample below) so the next health-cron and the Android/iOS UI know exactly what to surface.

```json
{
  "success": false,
  "code": "BOT_UNREACHABLE",
  "errorType": "connection_failed",
  "upstream": "bot_webhook",
  "retryable": true,
  "retry_after_ms": 30000,
  "botId": "mac_local_minimax_2_5",
  "botName": "Mac本地版_MiniMax2.5",
  "botKind": "free",
  "webhookHost": "mac-tunnel.example.com",
  "hint": "Could not reach the bot webhook. The bot host may be offline..."
}
```

## Test plan
- [x] `node --check backend/index.js`
- [x] `cd backend && npx jest tests/jest/official-borrow.test.js tests/jest/official-borrow-binding-reuse.test.js tests/jest/official-borrow-bind-502-structured.test.js --runInBand` → **38/38 pass**
- [x] New test `official-borrow-bind-502-structured.test.js` covers: every errorType → expected code/retryable (unit), and the wire path for `bind-free` returning the structured payload on `connection_failed` and on `HTTP 502` (integration via `supertest` + mocked global.fetch).

## What this does NOT fix
- The Mac local bot box still has to be up. This PR makes the failure mode legible to operators and to the UI; it does not repair an offline tunnel.
- `bind-personal` and `sync-binding` now return the structured payload in the same shape but were not explicitly covered by the new Jest test (they share the same helper so unit coverage transfers; runtime parity is mechanical).

## Refs
- GH#3001
- card_33679aa1113405512825c854

## Self-review (CLAUDE.md 10-item checklist)
- [x] Logic trace: handshake → errorType propagates → buildBindFailurePayload → 502 JSON
- [x] Test coverage: 9 new cases (7 unit + 2 supertest); both error shapes exercised
- [x] Return shape: backward compatible (kept `success:false`, `error`, `hint`; added new fields alongside)
- [x] What this does NOT fix: documented above
- [x] Security: no new auth surfaces, error messages do not leak secrets
- [x] /api/help: not affected (no new public endpoints)
- [x] Related docs: behavior described in PR body; no doc files need editing
- [x] i18n: no new user-facing strings — error remains zh; new fields are machine-readable codes
- [x] Info page: N/A
- [x] Debug endpoint: N/A — error response itself is the diagnostic

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -15844,11 +15844,14 @@ app.post('/api/official-borrow/bind-free', async (req, res) => {
     const handshake = await handshakeWithBot(freeBot.webhook_url, freeBot.token, preferredKey, deviceId, eId, 'free', freeBotAuthOpts);
 
     if (!handshake.success) {
-        return res.status(502).json({
-            success: false,
-            error: `無法與免費版機器人建立連線。${handshake.error || ''}`,
-            hint: 'Bot gateway may not have active sessions. Check bot configuration.'
+        const failurePayload = buildBindFailurePayload(handshake, {
+            botKind: 'free',
+            botId: freeBot.bot_id,
+            botName: freeBot.display_name || freeBot.bot_id,
+            webhookUrl: freeBot.webhook_url,
         });
+        failurePayload.error = `無法與免費版機器人建立連線 (${failurePayload.code}). ${handshake.error || ''}`.trim();
+        return res.status(502).json(failurePayload);
     }
 
     const sessionKey = handshake.sessionKey;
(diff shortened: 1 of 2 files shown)


HankHuang0516:
## CI status

**297/298 suites pass.** The single failing suite (`portal-smoke-static.test.js`) reports `dashboard.html` and `kanban.html` are missing `/shared-core/route-registry.js`. Verified: this test **also fails on `origin/main`** with the same two errors when the portal HTMLs from main and the test file from main are pulled in isolation. The failure is unrelated to this PR (which only touches `backend/index.js` bind-failure helpers + adds one new Jest file).

Proceeding to admin-merge per `## Self-review`. The 38 official-borrow tests including the 9 new structured-502 tests all pass locally.

```
Test Suites: 3 passed, 3 total
Tests:       38 passed, 38 total
```

PR #3496, opened 2026-06-18
[codex] fix chat embedding vector dimensions

HankHuang0516:
## Summary
- Normalize `chat_messages.embedding` to the checked-in `DEFAULT_DIM` (`vector(1536)`) before creating `idx_chat_embedding_hnsw`.
- Drop/recreate the HNSW index when repairing a legacy unbounded or wrong-dimension vector column, and clear incompatible generated vectors so they can be recomputed.
- Guard semantic search and async embedding writes so 1024-dimension provider output cannot poison the fixed-dimension pgvector schema.
- Make `backend/scripts/backfill-embeddings.js` run the same schema repair before recomputing `embedding IS NULL` rows.

## Root Cause / Decision
- Source inspection found no checked-in `backend/migrations` DDL for `chat_messages.embedding`; the live DDL is the runtime migration in `backend/chat-embedding.js`.
- `backend/embedding-client.js` (from commit `e523b29f`) defines the intended default as OpenAI `text-embedding-3-small` with `DEFAULT_DIM = 1536`.
- The only checked-in 1024 path is `CHAT_EMBEDDING_PROVIDER=voyage`, whose output is incompatible with the current fixed `vector(1536)` schema. Product/docs also describe OpenAI 1536 as the default.
- Therefore this PR treats 1024 output as the accidental/misconfigured path for the current schema, not as an intentional migration target. If EClaw later wants Voyage 1024 as the primary model, that needs a separate explicit schema/backfill plan.
- The prod `expected 384 got 1024` HNSW failure is consistent with an existing legacy/unbounded or wrong-dimension column/index plus newer 1024 vectors; `ADD COLUMN IF NOT EXISTS` could not repair that existing shape.

## Production Repair Path
1. Deploy this PR so startup normalizes the column/index and rejects wrong-dimension vectors.
2. Ensure the production embedding provider is the intended OpenAI 1536 path, not Voyage 1024, unless a new 1024 schema migration is intentionally planned.
3. Run `node backend/scripts/backfill-embeddings.js --max=0` with the production DB/API key context to recompute rows whose incompatible vectors were cleared to `NULL`.

Reviewer requested by EClaw card `card_563f12c86a8d6e7107747bf6`: #2 / LOBSTER.

## Validation
- `node -c backend/chat-embedding.js`
- `node -c backend/scripts/backfill-embeddings.js`
- `node -c backend/tests/jest/chat-embedding-unit.test.js`
- `npx jest tests/jest/chat-embedding-unit.test.js tests/jest/chat-search.test.js --runInBand`

### DIFF
--- a/backend/chat-embedding.js
+++ b/backend/chat-embedding.js
@@ -10,6 +10,42 @@ const { generateEmbedding, toPgVectorLiteral, DEFAULT_DIM } = require('./embeddi
 
 let vectorEnabled = false;  // set true after successful schema init
 let poolRef = null;
+const EXPECTED_VECTOR_TYPE = `vector(${DEFAULT_DIM})`;
+
+function vectorDim(vec) {
+    return Array.isArray(vec) ? vec.length : null;
+}
+
+function hasExpectedDim(vec) {
+    return vectorDim(vec) === DEFAULT_DIM;
+}
+
+async function normalizeEmbeddingColumn(pool) {
+    const result = await pool.query(`
+        SELECT pg_catalog.format_type(a.atttypid, a.atttypmod) AS column_type
+        FROM pg_catalog.pg_attribute a
+        WHERE a.attrelid = 'chat_messages'::regclass
+          AND a.attname = 'embedding'
+          AND NOT a.attisdropped
+        LIMIT 1
+    `);
+    const columnType = result.rows?.[0]?.column_type;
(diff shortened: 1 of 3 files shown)


HankHuang0516:
Superseded by #3497 (merged). Collision: I'd already merged your earlier working-tree version as #3495 before this clean PR finished; #3497 reconciles your unique backfill-embeddings.js schema-repair + refined guards onto main. Your work landed — thanks #6. Branch can be deleted.

PR #3583, opened 2026-06-20
fix(db): ensure fcm_token + apns_token columns at startup migration (card_caa6307)

HankHuang0516:
## Problem
`devices.fcm_token` and `devices.apns_token` were created **lazily** in `index.js` only when the first token of each platform was registered (`ALTER ... ADD COLUMN IF NOT EXISTS` with a swallowed `.catch(()=>{})`). On a fresh DB — e.g. the 2026-04 migration to `postgres-pgvector` — any query referencing `apns_token` before an iOS device ever registered errored with `column "apns_token" does not exist`.

Surfaced by #5's Railway PG log monitor; live schema diff confirmed `apns_token` was the only missing `devices` column on pgvector (`paid_borrow_slots`/`next_entity_id`/`prompt_policy`/`user_display_name`/`fcm_token` all present → no paid-feature impact).

## Fix
Move `ADD COLUMN IF NOT EXISTS` for both push-token columns into the `devices` startup-migration block in `db.js`, so the schema is guaranteed at boot regardless of push-registration traffic. Idempotent; no backfill needed.

Part of the card_caa6307 mobile-push cleanup (umbrella card_cff46d3). Low-impact (no iOS users yet) but removes recurring schema-drift errors per the no-benign-error principle.

### DIFF
--- a/backend/db.js
+++ b/backend/db.js
@@ -290,6 +290,21 @@ async function createTables() {
         `);
         console.log('[DynamicEntity] DB migration: next_entity_id column ensured on devices table');
 
+        // Push tokens (card_caa6307 follow-up): fcm_token (Android) + apns_token (iOS).
+        // Previously these were created lazily at runtime in index.js only when the FIRST
+        // token of each platform was registered (ALTER ... ADD COLUMN IF NOT EXISTS with a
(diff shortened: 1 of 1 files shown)


HankHuang0516:

PR #3594, opened 2026-06-20
fix(webpush): auto-enable on load when notification permission already granted (card_b3bb2bce)

HankHuang0516:
## Root cause (E2E-confirmed on prod)
Platform-wide **0 `push_subscriptions`**. Investigation + real-browser E2E showed the client subscribe flow, UI toggle, and VAPID **all exist and work** — but the "Enable browser notifications" toggle sits inside a **collapsed-by-default** notifications section, so users who had already granted notification permission never actually subscribed. Not missing code — a discoverability gap.

## Fix (per Hank: "通知區預設收合但默認啟動")
Keep the section collapsed, but make push **default-on**: `autoEnableWebPush()` runs on settings load and, when `Notification.permission === 'granted'` and there's no active subscription, silently registers the SW + subscribes + `POST /api/push/subscribe`.

**Browser-compliant:** never auto-prompts (permission requests need a user gesture), so `default`/`denied` are left to the manual toggle. Best-effort + silent; any failure leaves the manual flow intact. Reuses the same subscribe path as `toggleWebPush()`.

## Verify after merge
On a browser that already granted notification permission, load /portal/settings.html → a `push_subscriptions` row should appear without touching the toggle. (Per the lesson this card taught: confirm the actual artifact — the DB row + a delivered notification — not just code presence.)

### DIFF
--- a/backend/public/portal/settings.html
+++ b/backend/public/portal/settings.html
@@ -1811,6 +1811,7 @@
             initPassiveHealthCollapse();
             loadPassiveHealthSettings();
             updatePushToggle();
+            autoEnableWebPush(); // default-on: silently subscribe if permission already granted
 
             document.getElementById('pageContent').style.display = '';
             if (typeof window.applyDeepLinkFocus === 'function') window.applyDeepLinkFocus();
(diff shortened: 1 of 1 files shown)


HankHuang0516:

PR #3682, opened 2026-06-22
fix(wallpaper): resume watchdog + surface-valid draw gate; add Crashlytics (card_f9b2cc2d) [v1.1.3]

HankHuang0516:
## Card
card_f9b2cc2d — P0 Hank-direct, severe: live wallpaper goes ENTIRELY BLACK after a brief app-switch/return; only a full app kill recovers it.

## Root cause (residual black after #3650)
#3650 made `onSurfaceDestroyed` pause-only, but black persisted. The draw loop's reschedule gate read the `lifecycle.surfacePresent` **flag**, set true only by `onSurfaceCreated/Changed`. On a brief app-switch the framework can deliver `onVisibilityChanged(true)` **without** a paired `onSurfaceCreated` (surface re-validated late, or callback never re-fires) → flag stuck false → the 33ms loop never reschedules → permanent black until process kill.

## Fix (defensive, callback-order independent)
- **draw() gate** → uses authoritative `surfaceHolder.surface.isValid` instead of the `surfacePresent` flag, so the loop self-heals whenever visible + surface drawable.
- **Resume watchdog** → `onVisibilityChanged(true)`/`onSurfaceCreated`/`onSurfaceChanged` each fire a bounded retry burst (100/300/600/1200/2500ms) that kicks `draw()` as soon as the surface is valid, catching a missed/late surface callback. If the window elapses while visible but the surface never becomes valid → Crashlytics non-fatal (genuine stuck-black).
- Diagnostic loading markers retained (so a residual case still self-diagnoses).
- **Firebase Crashlytics** wired (same project as FCM — `eclaw-cc72a`; **no google-services.json change → push unaffected**) to capture stuck-black non-fatals in the field.

## Version
versionCode 112 → 113, versionName 1.1.2 → 1.1.3

## Verification
- `./gradlew :app:bundleRelease` → BUILD SUCCESSFUL, signed AAB produced.
- `:app:testDebugUnitTest` → 20/20 test files green (0 failures, 0 errors).
- Live-wallpaper resume behaviour requires on-device verification (per established native flow): merge → Play internal → Hank device-test "switch to launcher/another app and return → wallpaper repaints, no black".

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -6,6 +6,7 @@ plugins {
     alias(libs.plugins.kotlinAndroid)
     id("kotlin-kapt")
     alias(libs.plugins.googleServices)
+    alias(libs.plugins.firebaseCrashlytics)
 }
 
 android {
@@ -20,8 +21,8 @@ android {
         applicationId = "com.hank.clawlive"
         minSdk = 24
         targetSdk = 35
-        versionCode = 112
-        versionName = "1.1.2"
+        versionCode = 113
+        versionName = "1.1.3"
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
@@ -118,6 +119,11 @@ dependencies {
     // Firebase Cloud Messaging (FCM) for push notifications
     implementation(platform(libs.firebase.bom))
     implementation(libs.firebase.messaging)
+
+    // Firebase Crashlytics — captures the wallpaper resume black-screen
+    // non-fatals (card_f9b2cc2d). Same project as FCM (eclaw-cc72a); no
+    // google-services.json change, so push is unaffected.
+    implementation(libs.firebase.crashlytics)
 
     // Google Sign-In (Credential Manager)
     implementation(libs.credentials)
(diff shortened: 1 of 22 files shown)


HankHuang0516:
## 2nd commit (0cb4ca3) — hardened after Hank v1.1.2 device-test

Hank reproduced the black on v1.1.2 as **pure black with NO marker text**. Since markers paint in the lifecycle callbacks and the 33ms loop overwrites them, "no text" ⇒ loop alive painting black, or loop dead with no callback to restart. Both now handled:

- **draw() catch hardened**: records the real exception to Crashlytics (with entities/hasFirstResponse/surfaceValid keys) and overpaints a distinct dark-red "render error" frame (`claw_wallpaper_render_error`, 18 locales) instead of silently posting the half-drawn black canvas.
- **Self-healing heartbeat**: every ~1s, uses framework `isVisible` (not the cached flag); if we should be showing + surface valid but no frame posted for ~1s → kick draw(). Recovers in ~1s with no callback dependency (vs needing a full app restart).

Rebuilt: `:app:bundleRelease` BUILD SUCCESSFUL, `:app:testDebugUnitTest` 20/20 green.

PR #3709, opened 2026-06-23
fix(wallpaper): isolate TtsService in :tts process — kill the residual FGS-deadline black screen (card_f9b2cc2d) [v1.1.9]

HankHuang0516:
## Problem
The residual ~1/10 **"send msg → Home → ~3s → pure-black wallpaper, needs app restart"** survived every v1.1.3–1.1.8 in-process mitigation (render resilience, `startForeground` on every start, `START_NOT_STICKY`, FGS-crash swallow).

**Root cause:** a `startForegroundService()` that misses the OS ~5s `startForeground()` deadline is killed by the system via an **async** `ForegroundServiceDidNotStartInTimeException` (thrown by ActivityManager, *not* synchronously inside `startForeground()`), so the in-process try/catch in `TtsService.ensureForeground()` can never intercept it. While `TtsService` shared the **main** process, that kill took the live-wallpaper engine down with it → pure black until relaunch.

## Fix — process isolation (guaranteed, not just prevention)
- **AndroidManifest**: `TtsService` gets `android:process=":tts"`. The FGS now has its own idle main looper (makes the deadline) **and** any residual deadline-kill is scoped to the `:tts` process — the wallpaper in the main process survives.
- **ClawApplication.onCreate**: bail out after Timber when running in `:tts` (new `isTtsProcess()`/`currentProcessName()`, `getProcessName()` on P+ / `/proc/self/cmdline` on 24–27) so the heavy main-process bootstrap (Telemetry, FCM self-heal, Play Integrity, crash upload, channels) does **not** re-run / double-register in the isolated process.
- **SocketManager**: the socket `device:tts` event can no longer reach the now cross-process `TtsService` via the in-process `ttsFlow`, so forward each utterance with `startForegroundService(Intent + tts_text/lang/speed/pitch/entity)`, wrapped in try/catch so a backgrounded start can't crash the main process either.

## Verification (emulator Pixel_9a, vc119 debug)
- App boots clean: **main pid + `:tts` pid** both spawn; 0 FATAL / 0 `DidNotStartInTime`.
- Process guard fires: `:tts` logs `"running in :tts process — minimal init only"`; full init + FCM register only in main (no double-register).
- TTS not regressed: `[TTS] Engine initialized successfully` **inside the `:tts` process** (Google TTS cmn-TW loaded).
- **The fix, proven**: `run-as kill -9` of the `:tts` process (the exact thing the FGS-deadline kill does) leaves the **main process (wallpaper host) UNCHANGED and healthy** — FCM/WebView/Socket all still alive. Pre-isolation that same kill blacked the wallpaper.

## Notes
- Reopens card_f9b2cc2d (was prematurely closed at v1.1.8; the residual black screen was still reproducible — Hank reproduced it on the simulator).
- Version bump 118→**119** / 1.1.8→**1.1.9**. (Branch name says vc118 but the commit ships vc119.)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -21,8 +21,8 @@ android {
         applicationId = "com.hank.clawlive"
         minSdk = 24
         targetSdk = 35
-        versionCode = 118
-        versionName = "1.1.8"
+        versionCode = 119
+        versionName = "1.1.9"
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -189,9 +189,18 @@
             </intent-filter>
         </service>
 
+        <!-- card_f9b2cc2d: run TtsService in an isolated ":tts" process. A
+             startForegroundService() that misses the OS 5s startForeground()
+             deadline triggers an ASYNC system kill (ForegroundServiceDidNotStart-
+             InTimeException) that no in-process try/catch can intercept. While the
+             service shared the main process, that kill took the live-wallpaper
+             engine down with it (pure black, needs app restart). Isolating it gives
+             the FGS its own idle main looper (so it makes the deadline) AND scopes
+             any residual kill to the :tts process, leaving the wallpaper alive. -->
         <service
             android:name=".service.TtsService"
             android:exported="false"
+            android:process=":tts"
             android:foregroundServiceType="mediaPlayback" />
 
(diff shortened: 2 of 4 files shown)


HankHuang0516:
Closing per owner (Hank) decision 2026-06-28: 1.1.13 (vc124) already fixes the black screen with no remaining black-screen reports, so these residual-hardening PRs are being discarded for now. The 2026-06-28 update ships wallpaper settings + walking-behavior fixes. Branch is preserved and this PR can be reopened if a residual FGS-deadline / never-black case recurs.

PR #3746, opened 2026-06-25
[codex] Polish telemetry context panel

HankHuang0516:
## Summary
- add a live telemetry context/status panel with type-filter state and a clear type-filter action
- add refresh aria-busy/disabled state plus a request-sequence guard so stale telemetry responses cannot overwrite newer filters
- fix generated telemetry fallback labels so missing [html] keys do not render raw keys
- add focused static Jest coverage for the telemetry page UI

## Validation
- NODE_PATH=/Users/hank/Desktop/Project/EClaw/backend/node_modules /Users/hank/Desktop/Project/EClaw/backend/node_modules/.bin/jest --runTestsByPath backend/tests/jest/telemetry-page-ui.test.js backend/tests/jest/portal-static-ids.test.js --runInBand --testPathIgnorePatterns=/node_modules/ --testPathIgnorePatterns=/.claude/ --testPathIgnorePatterns=/.git/
- node backend/scripts/i18n-check.js (passes; existing soft settings-help fanout warnings remain)
- inline telemetry script parsed with espree
- git diff --check -- backend/public/portal/telemetry.html backend/tests/jest/telemetry-page-ui.test.js
- local Playwright mock smoke on http://127.0.0.1:8996/portal/telemetry.html?lang=en at desktop 1366x900, mobile 390x844, and split-pane 480x900: no console issues, no failed requests, no page-level horizontal overflow, clear-filter reset works
- screenshot QA: /tmp/eclaw-telemetry-ui-20260626/desktop.png, /tmp/eclaw-telemetry-ui-20260626/mobile.png, /tmp/eclaw-telemetry-ui-20260626/split-pane.png

## Computer Use
- Required browser Computer Use readback was attempted, but local approval denied both browsers: Chrome returned "The user did not approve Copilot Computer Use to use Google Chrome." Safari returned "The user did not approve Copilot Computer Use to use Safari." No Chrome/Safari title/URL readback could be performed.

## Safety
No production routes, real credentials, telemetry deletion, direct main push, self-merge, or production deploy were performed.

PR #3747, opened 2026-06-25
fix(chat): restore staged reply on send-abort paths, not just on catch (card_2625ae06)

HankHuang0516:
## What
`sendMessage()` runs `clearReplyContext()` **optimistically** before two early-return guards:
1. uploads-still-pending (`chat_wait_upload`)
2. no routing target selected (`chat_select_entity`)

The snapshot-restore that puts the quote chips back lived **only in the `catch` block**, which those two `return`s never reach. So a send aborted on either path **silently discarded** the user's staged quote chips **and** their action-request reply binding — the message never sent and the linkage to the「需要你」inbox item was orphaned, with no feedback.

## Fix
- Extract the catch-block restore into a DRY `restoreReplyContextsOnAbort()` closure (defined right after the existing `replyContextsSnapshot`).
- Invoke it on **both** early-return guards, and reuse it in the `catch` (no behavior change to the catch path).
- Net: every abort path now restores the staged reply; the happy path is untouched (`clearReplyContext()` still consumes the quotes on a real send).

## How found
Daily QA/UIUX regression sweep (card_2625ae06), run while covering rate-limited #1. Med-Low severity, narrow blast radius (needs uploads-pending OR no-target while a quote/action-request is staged) but it's silent data loss.

## Test hardening (same PR)
`chat-targets-static.test.js` sliced `sendMessage` with a **fixed 6000-byte window** to assert the no-target guard. The new helper (~780 chars near the top of the function) pushed the **intact** guard past 6000 — the exact fragility that previously forced a 5000→6000 widen (card_277c80f5). Switched the slice to a **semantic end-marker** (`btnSend` lookup immediately after the guard) so the assertion is shift-proof and won't recur.

## Tests
- `chat-action-request-inbox.test.js` — added an assertion that all three abort paths (uploads / no-target / catch) restore via the helper.
- `chat-action-request-inbox` + `chat-targets-static`: **19/19 green** locally.
- (Local full-suite run was in a deps-less worktree; authoritative full green is this PR's CI.)

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -8486,6 +8486,18 @@
             // restore them (and their action-request binding) instead of orphaning
             // the request (card_b176c435).
             const replyContextsSnapshot = replyContexts.slice();
+            // Restore the optimistically-cleared quote chips + their action-request
+            // binding on ANY abort path — the early-return guards below (uploads still
+            // pending / no routing target) as well as the catch. Without this a send
+            // that never leaves the client silently orphans the staged reply with no
+            // feedback (card_2625ae06; extends the catch-only restore of card_b176c435).
+            const restoreReplyContextsOnAbort = () => {
+                if (replyContextsSnapshot.length && replyContexts.length === 0) {
+                    replyContexts = replyContextsSnapshot.slice();
+                    renderReplyPreviews();
+                    recomputeAutoReceivers();
+                }
+            };
             let finalText = text;
             if (replyContexts.length) {
                 // One "↩ …" line per quote so the recipient entity receives EVERY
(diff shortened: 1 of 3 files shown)


HankHuang0516:
CI fully green (Jest unit tests pass, Required PR CI gate pass, matrix pass, ESLint/i18n/critical-pages pass). MERGEABLE.

**Held for the hard UX vision-gate** (this changes user-visible chat.html behavior — staged reply chips persist on aborted send). Self-serving the Playwright check needs the vault DEVICE_SECRET, which requires Hank's explicit authorization. So this is batched into the pending portal-vision authorization (with #3746 + 需要你 inbox close-out card_b176c435), then merged. Not auto-merging.

PR #3817, opened 2026-06-30
fix(petdx): proxy community sprites same-origin to defeat Chrome ORB 🦞 fallback

HankHuang0516:
## Problem — every community partner shows the 🦞 lobster fallback

The Companion Browser (`petdx-browser.html`) rendered the default **🦞 lobster fallback for every community partner**.

**Root cause (verified, not re-litigated):** community/partner pets whose sprite bytes failed to self-host in R2 keep the **external** CDN URL `https://assets.petdex.dev/pets/<slug>/sprite.webp`. Chrome **ORB (Opaque Response Blocking)** blocks that cross-origin `.webp` in the browser — `curl` gets `200` (ORB is browser-only), but the browser logs `net::ERR_BLOCKED_BY_ORB`. Setting `crossOrigin='anonymous'` does **not** help (the CDN's `Access-Control-Allow-Origin` is inconsistent). The existing same-origin proxy `/api/petdx/:slug/sprite.webp` only covers the R2-hosted `petdx-sprites/` objects, not these external community sprites.

## Approach ② — proxy the external CDN same-origin (owner-decided)

Since `assets.petdex.dev` is external, we stream it through our own origin. A same-origin response is outside ORB's scope, so the sprite loads.

### Backend proxy — `petdex-route.js`
New route `GET /api/petdx/community/:slug/sprite.webp` fetches the upstream bytes and streams them back with `Content-Type: image/webp` and a 1-day `Cache-Control` (not `immutable` — community sprites can recover/update upstream).

**SSRF guard — this is NOT an open proxy:**
- The upstream host and the `/pets/<slug>/sprite.webp` path shape are **hard-coded literals**.
- Only a **charset-validated slug** (`/^[a-z0-9_-]+$/i`) is interpolated.
- No host, scheme, or path is ever taken from the request, so it cannot be coerced into requesting an arbitrary URL. Bad slugs (dots, encoded slashes, traversal) are rejected with `404` **before** any fetch.

### Serializer rewrite — `companion-api.js` + `petdex-bridge.js`
`rowToCompanionCard` / `rowToCompanionDetail` rewrite the external sprite URL to the proxy path on `asset_url`, `avatar_url`, **and** `descriptor.asset.url` (the URL the renderer actually loads, and the one the detail modal renders from). The rewrite helpers (`rewriteCommunitySpriteUrl` / `rewriteDescriptorSpriteUrl`) are a clone-safe **no-op** for already-proxied / R2-hosted / null URLs and for any other host or path.

### Grid `kind` fallback — `petdx-renderer.js` + `petdx-browser.html`
The grid's synthetic descriptor omitted `kind`, so the emoji fallback always defaulted to 🦞 even for character/object companions. The serializer now surfaces `kind`, the browser passes `kind`+`category` into the synthetic descriptor, and `fallbackEmojiForKind` (extracted + exported) infers `kind` from `category` so a load failure draws 🐾 / 🧑 / 🎯 by kind.

### Load-failure sort demotion — `petdx-browser.html`
Under 熱門 / 最新, a sprite that still fails to load is **stably demoted** below cards that render (probe `Image` + a secondary sort key on a load-ok flag), so the first rows aren't all broken avatars. The server's primary sort order is preserved within each group; rating/favorites keep their exact server order.

## Test evidence

New `backend/tests/jest/petdx-community-sprite-proxy.test.js` (16 tests):
- **(a)** proxy accepts the valid `assets.petdex.dev/pets/<slug>/sprite.webp` shape (asserts the locked-down upstream URL) and **rejects** other hosts/paths/slug charsets with no fetch (SSRF guard); upstream 404→404, upstream error→502.
- **(b)** serializer rewrites the external URL → proxy path on card + detail, surfaces `kind`, never mutates the source descriptor, and is a no-op for proxied/R2 URLs.
- **(c)** `fallbackEmojiForKind` returns non-lobster for character/object/creature and infers kind from category.

```
Tests:       16 passed (new file)
Tests:       162 passed (existing petdx/companion suites — no regression)
```

## Caveat to verify in a real browser
Tests mock the upstream `fetch`; they prove the SSRF guard, URL rewrite, and kind fallback, but not the live ORB behavior. Worth a quick prod check that `GET /api/petdx/community/<slug>/sprite.webp` returns the bytes `200` (the upstream `assets.petdex.dev` accepts the `EClaw-petdex-proxy/1.0` User-Agent — if its edge filter is UA-strict, swap to a curl-style UA) and that a community card now renders the real sprite (or its per-kind emoji on a genuine miss) instead of 🦞.

### DIFF
--- a/backend/companion-api.js
+++ b/backend/companion-api.js
@@ -18,7 +18,10 @@ const express = require('express');
 const fs = require('fs');
 const path = require('path');
 const { Pool } = require('pg');
-const { syncPetdexCatalog, R2_KEY_PREFIX, PROXY_PATH_PREFIX } = require('./petdex-bridge');
+const {
+    syncPetdexCatalog, R2_KEY_PREFIX, PROXY_PATH_PREFIX,
+    rewriteCommunitySpriteUrl, rewriteDescriptorSpriteUrl,
+} = require('./petdex-bridge');
 const { extractFrameZero } = require('./companion-avatar-util');
 const extractCreds = require('./extract-creds');
 
@@ -162,7 +165,9 @@ function rowToCompanionCard(row) {
         id: row.id,
         name: row.name,
         version: row.version,
-        avatarUrl: row.avatar_url,
+        // Rewrite external community-CDN URLs to the same-origin proxy so the
+        // browser doesn't ORB-block them (no-op for R2/proxy/null URLs).
+        avatarUrl: rewriteCommunitySpriteUrl(row.avatar_url),
         thumbnailUrl: row.thumbnail_url,
         author: row.author_entity_id != null
             ? { entityId: row.author_entity_id }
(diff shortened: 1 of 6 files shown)


HankHuang0516:
Superseded by #3888 (resurrected onto current main + real-render verified: same-origin proxy defeats Chrome ORB, proxied sprite naturalWidth=1536 vs external=0). This PR was never merged and is 39 commits stale.

PR #3900, opened 2026-07-04
fix(portal): make chat key refs server-authoritative

HankHuang0516:
## Summary
- Make chat Key Reference read /api/device-vars with GET only; remove the background POST merge that could resurrect stale localStorage keys.
- Reconcile chat localStorage from the server response, removing stale cached keys when the server vault no longer has them.
- Hide/clear Key Reference chips when the server vault is empty or unavailable instead of showing stale cache.

## Validation
- npx jest tests/jest/chat-env-var-reference.test.js --runInBand
- npm run smoke:portal

Card: card_f6038f0fe7ef8f07d2741c3e

### DIFF
--- a/backend/public/portal/chat.html
+++ b/backend/public/portal/chat.html
@@ -10137,37 +10137,37 @@
         // ── Env Var Reference ──
         async function initEnvVarBtn(deviceId, deviceSecret) {
             const storageKey = `eclawLocalVars_${deviceId}`;
-            let vars;
-            try { vars = JSON.parse(localStorage.getItem(storageKey) || '{}'); } catch { vars = {}; }
+            const envBtn = document.getElementById('attachMenuEnvBtn');
+            const chipsEl = document.getElementById('envVarChips');
+            let vars = {};
 
-            // Sync from server so WebView / new sessions pick up vars set elsewhere.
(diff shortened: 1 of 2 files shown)


HankHuang0516:
Superseded by #3901 (merged) — thanks #1, your approach was right and I adopted two of your ideas into the merged fix: `encodeURIComponent` on the device-vars GET and the `Array.isArray` guard.

Why #3901 over this one (objective, not territorial):
1. **Green CI.** This PR's Jest went red because it removed chat.html's device-vars POST but didn't update `portal-vault-post-sends-etag.test.js`, which hard-asserts chat.html contains `expectedUpdatedAt`. #3901 updates that guard (env-vars.html keeps the etag check; chat.html is pinned to *not* POST the vault at all).
2. **Full cache reconcile.** This PR displays server truth but leaves a partially-deleted key lingering in `eclawLocalVars` (inert). #3901 rewrites the cache to match the server on every GET (acceptance #2 in full).
3. **Wider regression test** (deletion-propagates + no-POST, fresh session, empty vault, offline fallback) — fails on the old POST-merge code, passes on the fix.

Same root fix, same intent. Closing to avoid a double-merge on the same function.

PR #4817, opened 2026-09-10
fix(aihankapps): version immutable report data

HankHuang0516:
## Summary
- load AiHankApps report data through a content-hash URL so Cloudflare immutable caching cannot serve a previous report
- publish a newly validated report bundle with matching release hashes
- keep the release generator durable and add regression coverage for first-time and replacement cache-busting

## Verification
- 60/60 report pipeline tests pass
- published report release integrity validation passes
- strict i18n reference check passes
- i18n syntax suite: 12/12 pass
- `git diff --check` passes

## Production issue reproduced
The public `release.json` was current while unversioned `data.js` returned the previous immutable Cloudflare object. Adding a content-hash query retrieved the current deployment; this PR makes that URL part of every generated report release.


HankHuang0516:
## Codex self-review

Review target: `origin/main...8b777feb2` (5 files, cache-busting generator + generated report artifacts).

### Findings
- **No blocking findings.** The report model and release metadata use the same content hash, and `release.json` hashes validate all three public artifacts.
- The URL rewrite is fail-closed: release preparation stops if the report index no longer contains the expected `./data.js` script.
- The replacement is idempotent and updates both an unversioned URL and an older versioned URL.
- No authentication, user data, API routes, secrets, or localization keys are changed.

### Verification
- 60/60 reporting regression and integrity tests passed.
- Added regression cases for first-version, replacement-version, and missing-script failure.
- i18n syntax: 12/12 passed; strict i18n check scanned 118 HTML + 69 JS files with all EN references resolving and 0 simplified-character offenders.
- `git diff --check` passed.

### Residual risk / rollout check
- Cloudflare may retain the old unversioned object for its TTL, but this release no longer requests it; after deployment I will verify the exact hash-versioned URL and compare both public portfolio surfaces byte-for-byte for the four required report artifacts.

